PR #10, opened 2019-06-12
More useful example

ai:
This example will be more useful for most of the users.

But I am not sure how it will work in a branch. Will it compare `current_branch` vs `origin/current_branch`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ pre-commit:
 ## Custom file list
 
 Lefthook can be even more specific in selecting files.
-If you want to choose diff of all changed files between the current branch and master branch you can do it this way:
+If you want to choose all files changed locally but pushed to GitHub yet.
 
 ```yml
 # lefthook.yml
@@ -133,7 +133,7 @@ If you want to choose diff of all changed files between the current branch and m
 pre-push:
   commands:
     frontend-style:
-      files: git diff --name-only master # custom list of files
+      files: git diff --name-only origin # custom list of files
       glob: "*.{js}"
       run: yarn stylelint {files}
 ```
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ pre-commit:
 ## Custom file list
 
 Lefthook can be even more specific in selecting files.
-If you want to choose all files changed locally but pushed to GitHub yet.
+If you want to choose all files changed locally but not yet pushed to GitHub.
 
 ```yml
 # lefthook.yml


README.md
```diff
@@ -125,15 +125,15 @@ pre-commit:
...
+If you want to choose all files changed locally but pushed to GitHub yet.
 
 ```yml
 # lefthook.yml
 
 pre-push:
   commands:
     frontend-style:
-      files: git diff --name-only master # custom list of files
+      files: git diff --name-only origin # custom list of files
```
HellSquirrel: I think this command is the same as the previous one. How about `git diff @{upstream}`?

ai: `master` === local master. `origin` means GitHub.

I didn’t test `@{upstream}` (and remove Lefhook) can you test it?

HellSquirrel: Sure!

HellSquirrel:
```suggestion
      files: git diff --name-only @{push} || git diff --name-only master # custom list of files
```

README.md
```diff
@@ -125,15 +125,15 @@ pre-commit:
 ## Custom file list
 
 Lefthook can be even more specific in selecting files.
-If you want to choose diff of all changed files between the current branch and master branch you can do it this way:
+If you want to choose all files changed locally but pushed to GitHub yet.
```
HellSquirrel:
```suggestion
If you want to choose all files changed locally but not yet pushed to GitHub.
```

README.md
HellSquirrel: This command has fallback to master. Tested in local console :)

ai: Do we need a fallback to master? My suggestion:

1. If we are in `master` we need to compare with `origin/master`
2. If we are in `feature` we need to compare with `origin/feature`

Arkweid: Command `git diff --name-only origin` return a error.
I suppose `git diff --name-only @{push}` enough for that kind example?

HellSquirrel: I'm not sure. Sometimes there is no upstream for the current branch (for example `git checkout -b`). In such a case the check will always fail :(

cc @Arkweid

Arkweid: I'm not sure that we need something complicated in this example.
I think it is enough to make it clear that you can throw any command on this line, so long as it returns the list of files.

ai: Yeap, maybe you are right and we have a bigger problem here.

What if we change the docs:
1. We will start with a real useful copy-paste example.
2. In the next chapters, we will explain how to create complicated things.

This issue is about my problem: I have `logux/logux` project where I want to use Lefthook and I do not know what to write in config. The most popular thing: check all changed files before `git push`. The current docs don’t tell me how to do it. We need to improve them to make migration easier.

PR #35, opened 2019-07-31
Add links to the post

palkan:
/cc @ai


ai:
Quote description doesn't work well because it is low contrast and was not written according the user questions.

However, the definitely should have the link to the article.

PR #36, opened 2019-07-31
Flag to force or disable colored output

### LINKED ISSUE #33: Add option to disable a colors
For now we disable a colors if output a not a terminal.
Add a top level option
```
colors: false
```
And check it here:
https://github.com/Arkweid/lefthook/blob/0f8826ac9e85532143b8dec3256f5552d8c19745/cmd/root.go#L107-L109

orsinium:
1. Disable colored output if `--no-colors` flag specified
1. Force colored output if `colors: true` specified in the config.
1. Disable colored output if `colors: false` specified in the config.
1. Use old behavior by default: enable colors only for TTY
1. `IsTTY` replaced by `EnableColors`, because now it's about colors, not TTY.

Close #33 

Should I update some docs or do some tests? I haven't noticed a good place for it in the current codebase.

Thank you for the nice tool!

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -52,10 +52,16 @@ func Execute() {
 }
 
 func init() {
+	initAurora()
 	cobra.OnInitialize(initConfig)
-	au = aurora.NewAurora(IsTTY())
+	// re-init Aurora after config reading because `colors` can be specified in config
+	cobra.OnInitialize(initAurora)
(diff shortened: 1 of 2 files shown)


ai:
Should we take some checks from the JS library to detect color support? https://github.com/chalk/supports-color/blob/master/index.js

Arkweid:
Great work @orsinium! Thank you for contribution :)

PR #40, opened 2019-08-01
Add --no-pager flag to brakeman

jaydorsey:
I followed the configuration/setup for Ruby, and have the following lefthook.yml config:

```
---
pre-commit:
  commands:
    brakeman:
      run: bundle exec brakeman
```

I noticed that brakeman "gets stuck":

image

I tried a number of flags (parallel, output, debug, bundler, no bundler, etc.) but I think the cause is the default paging behavior which the brakeman docs note [here](https://brakemanscanner.org/docs/options/):

> By default, Brakeman pages output to the terminal with the less pager. To have Brakeman output directly to terminal, use `brakeman --no-pager`

I thought this might be a useful improvement to the docs in case anyone else runs into this

I'm running macOS Mojave, brakeman 4.6.1 (via gem install), lefthook 0.6.3 (via gem install), ruby 2.6.3

### DIFF
--- a/docs/other.md
+++ b/docs/other.md
@@ -34,7 +34,7 @@ pre-commit:
   parallel: true
   commands:
     audit:
-      run: brakeman
+      run: brakeman --no-pager
     rubocop:
       files: git diff --name-only @{push}
       glob: "*.{rb}"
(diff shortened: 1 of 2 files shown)


Arkweid:
Thank you @jaydorsey. This is a good point.

Issue #16, opened 2019-06-14
Speed up on Node


ai:
Now we have 3 binaries in npm and [script](https://github.com/Arkweid/lefthook/blob/master/.npm/bin/index.js) which run necessary binary depends on the current platform.

There are 2 big problems with current behavior:

1. npm package size is huge (26 MB)
2. Node.js startup time is not great.

Solution:

1. `postinstall` script download necessary binary (many npm packages use this approach, for instance `puppeter`).
2. npm runs binary directly without `index.js`

ai:
On CI we will avoid even downloading binary

fleischie:
Hello all, especially @HellSquirrel (as you assigned this task to you),

How do I ask politely if I can offer my help with this task?

<details><summary>I made a prototype of how to download the binary on postinstall</summary>
I made a crude mockup that downloads the appropriate binary for the current OS (or fail if the OS isn't supported), and afterwards run the `install` command with the downloaded binary. The postinstall script already had a check for CI in place, that can be used to skip the downloading in it's entirety.

I am unsure still how to handle errors on downloading the binary.
</details>

But I wanted to ask first, whether this task is still being developed on, and whether I should share my version with somebody.

HellSquirrel:
Hi @fleischie ! Thank you for your question and interest in this feature. 
I planned to do this task later, but if you want to help us it would be great!
If you already have a working draft just open a new pull request and we will continue the discussion there.
> Also I don't know where to store the binaries, if not in the .npm directory?

ai:
On network error we should try 3 times and print error message with 1 exit code if we still have network error.

Also we need environment variable to prevent downloading (for instance, if there too is not used on CI).

ai:
Puppeteer puts binaries to `node_modules/puppeteer` dir. We can do the same and put them in `lefthook` dir.

sHooKDT:
I have a couple of questions here:

1. This way we break the possibility to install the package without the network. There are projects where CI runs in isolated environment and uses e.g. yarn offline mirror to achieve non-network-bound build process. I think we should at least provide a way to manually download and store the needed binary in project directory.
2. What if someone develops in mixed environment, e.g. WSL + Windows? It's not that rare when you are running dev server under WSL, but make a commit under Windows. Can we provide a way to download both needed binaries?

Sorry for skepticism, but the current solution covers such cases with zero effort.

ai:
1. Isolated environments can use non-npm Lefthook distribution.
[...]

PR #43, opened 2019-08-02
Download binary on npm postinstall hook

fleischie:
As was discussed in issue #16 the `lefthook` npm package contains all the binaries, this postinstall hook downloads only the appropriate binary for the host OS (or in case of CI: none).

Note: I submit my PR in accordance to https://github.com/Arkweid/lefthook/issues/16#issuecomment-517672355.
Note 2: I did not run any go/ruby command as this PR mainly changes the postinstall script in the `.npm` directory. If this is still necessary I will update my PR, just let me know what is missing. :v:

### DIFF
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -48,7 +48,7 @@ function resolveDownloadPath(os, version, arch) {
 function downloadBinary(os, callback) {
   // construct single binary file path
   const binaryDir = join(packageDir, "node_modules", "@arkweid", "lefthook", "bin");
-  const binaryPath = join(binaryDir, binaryName);
+  const binaryPath = join(binaryDir, `${binaryName}${os === "win32" ? ".exe" : ""}`);
 
   // create directory if not existing, yet
   try {
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -1,6 +1,6 @@
 require("process");
 const { spawn } = require("child_process");
-const { createWriteStream, unlink } = require("fs");
+const { createWriteStream, mkdirSync, unlink } = require("fs");
 const { createGunzip } = require("zlib");
 const { get } = require("https");
 const { join } = require("path");
@@ -42,7 +42,18 @@ function resolveDownloadPath(os, version, arch) {
 
 function downloadBinary(os, callback) {
   // construct single binary file path
-  const binaryPath = join(process.env.INIT_CWD, "node_modules", "@arkweid", "lefthook", "bin", binaryName);
+  const binaryDir = join(process.env.INIT_CWD, "node_modules", "@arkweid", "lefthook", "bin");
+  const binaryPath = join(binaryDir, binaryName);
+
+  // create directory if not existing, yet
+  try {
+    mkdirSync(binaryDir);
+  } catch (err) {
+    if (err.code !== 'EEXIST') {
+      console.error(err.message);
+      process.exit(0);
+    }
+  }
 
   // create write stream as file with the constructed binaryPath, set it
   // up to close as soon as no more data is expected
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -3,11 +3,12 @@ const { spawn } = require("child_process");
 const { createWriteStream, mkdirSync, unlink } = require("fs");
 const { createGunzip } = require("zlib");
 const { get } = require("https");
-const { join } = require("path");
+const { join, resolve } = require("path");
 
 const { version } = require("./package.json");
 
 const binaryName = "lefthook";
+const packageDir = process.env.INIT_CWD || resolve(".");
 
 function closeAndRemoveFile(file) {
   file.on("close", () => {
@@ -42,7 +43,7 @@ function resolveDownloadPath(os, version, arch) {
 
 function downloadBinary(os, callback) {
   // construct single binary file path
-  const binaryDir = join(process.env.INIT_CWD, "node_modules", "@arkweid", "lefthook", "bin");
+  const binaryDir = join(packageDir, "node_modules", "@arkweid", "lefthook", "bin");
   const binaryPath = join(binaryDir, binaryName);
 
   // create directory if not existing, yet
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -26,13 +26,17 @@ function resolveDownloadPath(os, version, arch) {
   } else if (os === "linux") {
     osStr = "Linux";
   } else if (os === "win32") {
+    if (arch === "ia32" || arch === "x86") {
+      console.error("Lefthook is only supported on Windows with 64bit architecture.");
+      process.exit(0);
+    }
     osStr = "Windows";
   }
 
-  if (arch === "ia32" || arch === "x86") {
-    archStr = "i386";
-  } else if (arch === "x64") {
+  if (arch === "x64") {
     archStr = "x86_64";
+  } else if (arch === "ia32" || arch === "x86") {
+    archStr = "i386";
   } else {
     console.error(`Unable to resolve architecture (${arch}), only supported architectures are: 'ia32', 'x86', and 'x64'`);
     process.exit(0);
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -1,23 +1,55 @@
 require("process");
 const { spawn } = require("child_process");
+const { createWriteStream, unlink } = require("fs");
+const { get } = require("https");
 const { join } = require("path");
+
+const downloadPaths = {
+  "darwin": "https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-mac",
+  "linux": "https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-linux",
+  "win32": "https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-win.exe"
+};
+
+function downloadBinary(os, callback) {
+  // construct single binary file path
+  const binaryPath = join(process.cwd(), 'node_modules', '@arkweid', 'lefthook', 'bin', binary);
+
+  // create write stream as file with the constructed binaryPath, set it
+  // up to close as soon as no more data is expected
+  const file = createWriteStream(binaryPath, { mode: 0o755 });
+  file.on("finish", () => {
+    file.close();
+  });
+
+  // start download of binary file, set it up to remove the file if an
+  // error occurs
+  const download = get(downloadPaths[os], response => {
+    response.pipe(file);
+    response.on("end", () => {
+      callback(binaryPath);
+    });
+  });
+  download.on("error", err => {
+    console.error("Unable to download lefthook binary file");
+    console.error(err.message);
+    unlink(file);
+  });
+}
+
+function installGitHooks(binaryPath) {
+  console.log("Installing git hooks");
+  spawn(binaryPath, ["install", "-f"], { stdio: [process.stdin, process.stdout, process.stderr] });
+}
 
 const isCI = process.env.CI;
 if (!isCI) {
   process.chdir(process.env.INIT_CWD);
 
-  if (process.platform === 'darwin') {
-    binary = 'lefthook-mac';
-  } else if (process.platform === 'linux') {
-      binary = 'lefthook-linux';
-  } else if (process.platform === 'win32') {
-      binary = 'lefthook-win.exe';
-  } else {
+  if (!["darwin", "linux", "win32"].includes(process.platform)) {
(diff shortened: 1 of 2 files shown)


Arkweid:
/cc @HellSquirrel 
#16

HellSquirrel:
Thanks! I will test it in a few days

.npm/postinstall.js
```diff
@@ -1,23 +1,55 @@
 require("process");
 const { spawn } = require("child_process");
+const { createWriteStream, unlink } = require("fs");
+const { get } = require("https");
 const { join } = require("path");
 
+const downloadPaths = {
+  'darwin': 'https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-mac',
```
HellSquirrel: This script will always install the latest release from the 'master' branch. 
This will cause problems if we want to do something like `npm i @arkweid/lefhook@1.0.0`

.npm/postinstall.js
```diff
@@ -1,23 +1,55 @@
...
+  'linux': 'https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-linux',
+  'win32': 'https://raw.githubusercontent.com/Arkweid/lefthook/master/.npm/bin/lefthook-win.exe'
+};
+
+function downloadBinary(os, callback) {
+  // construct single binary file path
+  const binaryPath = join(process.cwd(), 'lefthook');
```
HellSquirrel: `process.cwd()`  is our current working directory. Consider some examples
`npm i @arkweid/lefhook` -- it will upload the binary to the project root
`cd src && npm i @arkweid/lefhook` -- it will upload the binary to 'src' 
We want the binary to be uploaded to `./node_modules/@arkweid/lefhook/...`

.npm/postinstall.js
fleischie: Yes. That's bad.

Would you suggest to resolve the packages from the releases/tags URL? `https://github.com/Arkweid/lefthook/releases/download/${version}/lefthook_${version}_${os}_${arch}.gz`?

That raises three more questions:
- How to get the installed version from the postinstall script,
- how to determine the `arch` value, and
- whether to unpack the binary from the package via the postinstall script?

.npm/postinstall.js
fleischie: Oh, whoops. I guess that was just the leftover `binaryPath` from my local tests. Will update immediately.

fleischie: `process.cwd()` should also give us the base path of the project, as the main function calls `process.chdir(process.env.INIT_CWD)`.

But if you want I can rewrite this, so that both the `downloadBinary` function takes `process.env.INIT_CWD` into consideration for the path, as well as call `process.chdir(process.env.INIT_CWD)` inside `installGitHooks`.

.npm/postinstall.js
fleischie: OOooohh, and trying out some stuff: How to deflate the windows' zip-files with node? I don't really want to include a `zip`-related node-package just for that specific reason...

fleischie: And another thing I just noticed: GitHub doesn't serve the files directly, but need to follow the redirect that the releases-binaries responses contain.

I'm not saying it's impossible, I'll just need more time to implement a proper solution.

Sorry for spamming, I'm just writing as I go. 😛

fleischie: I'm having troubles finding a good solution for windows: I could rely on just downloading the file to a temporary location, call the (hopefully) installed `zip` binary (I would have to check how to do that on a windows machine) and then remove the zip-package.

This of course is a bit more complicated than the MacOS/Linux version, which only needs to deflate the downloaded `.gz` via the node-internal `zlib` module (and which I have a working prototype for).

HellSquirrel: @Arkweid what do you think? :)
I think we should put [zlib-compatible](https://nodejs.org/api/zlib.html) binaries to the releases folder. 
As for redirects, you can do something like
```
if (response.statusCode === 302) {
       //get a new url from response.headers.location
   } else {
      // load file
};
```

.npm/postinstall.js
HellSquirrel: We have two problems:
1. `process.env.INIT_CWD` refers to the current folder. (It could be a subfolder of the root folder.) So we need to look for `node_modules` recursively
2. Some older versions of `yarn`  don't set this variable at all

HellSquirrel:
Here how you can load unzip and execute lefthook binary on mac
```const https = require("https");
const fs = require("fs");
const zlib = require("zlib");

const url =
  "https://github.com/Arkweid/lefthook/releases/download/v0.6.3/lefthook_0.6.3_MacOS_x86_64.gz";

const fileName = "lefthook-local";

https
  .get(url, res => {
    if (res.statusCode === 302) {
      const originalUrl = res.headers.location;
      console.log("redirect, redirect");

      https
        .get(originalUrl, res => {
          if (res.statusCode === 200) {
            res
              .pipe(zlib.createGunzip())
              .pipe(fs.createWriteStream(fileName))
              .on("finish", res =>
                fs.chmod(fileName, 755, err => {
                  if (err) {
                    console.log("error changing rights");
                  }

                  console.log("done");
                })
              );
          }
        })
        .on("error", e => {
          console.log("error after redirect!", e);
        });
    }
  })
  .on("error", e => {
    console.error(e);
  });
```

HellSquirrel:
For Windows we should change the `url` variable to `https://github.com/Arkweid/lefthook/releases/download/v0.6.3/lefthook_0.6.3_Windows_x86_64.gz`

.npm/postinstall.js
levenkov: 1. No, you shouldn't search the folder node_modules at all (e.g., name of "node_modules" folder can be changed). There is simple and steady solution: https://github.com/levenkov/lefthook/pull/1/files#diff-5fe0be6c71fdcee66beee780a898073fR167

1. Download it in preinstall hook.
2. preinstall hook executed in the package directory.
3. It's always true, that if you placed downloaded binary in `./bin` in the preinstall hook, this binary after install stage will be in the right place (usually here: `<project dir>/node_modules/@arkweid/lefthook/bin/...`).

I almost done the work, but was overwhelmed by the deadline :)

fleischie: I was under the impression that the `INIT_CWD` is set to the base path of the newly installed package (see https://github.com/yarnpkg/yarn/blob/cfb560e1b2e37ed5cfe5d2374a806ac3ae512c11/src/util/execute-lifecycle-script.js#L40 similar for `npm`), hence `INIT_CWD` for `lefthook` would always be `<project-dir>/node_modules/@Arkweid/lefthook`.

(Also the change was (according to GitHub) over a year ago, so I would assume most users have a current enough `yarn`/`npm` installation)



So tl;dr: As long as we don't `process.chdir` or `cd <somewhere> && ...` in the postinstall script, we should be able to assume we are in `<project-dir>/node_modules/@Arkweid/lefthook`, no?

fleischie:
(I will have time (maybe today but definitely) on thursday to continue working on this PR. Sorry for the inconvenience. 🙇 )

.npm/postinstall.js
levenkov: > (Also the change was (according to GitHub) over a year ago, so I would assume most users have a current enough yarn/npm installation)

It's better to support as earlier as you can version of node. It's common place to support >=6.0.

INIT_CWD is working after 8.

.npm/postinstall.js
fleischie: @HellSquirrel I updated the PR, it should now fetch the binaries from the releases paths (with redirect-resolve). I was only able to test on linux-x64, but it seems to work alright. (I also adapted the bin-path creation from @levenkov, which makes sense if we not deliver all the binaries in the npm package.)

I will click the "Resolve conversation" button now, because I feel this particular discussion is resolved, feel free to open a new one on the updated `postinstall.js`.

.npm/postinstall.js
fleischie: @levenkov @HellSquirrel 

I honestly don't know how to fix this `INIT_CWD` issue. I tried to use the `./dir` approach (i.e. `const path = process.env.INIT_CWD || path.resolve(".");`), but in my locally installation tests, this always resolved to the path the script resides in (i.e. `<path>/lefthook/.npm`, instead of `<path>/tmp` where I called `npm install ../lefthook/.npm` from).

Maybe you can review the latest PR version and give me some pointers.

levenkov: @fleischie I'll see that in 1-2 days. Maybe I'll try to change the approach in your code to mine and suggest small patch. I don't claim to replace all your work, just this problem.

ai:
Is it ready?

fleischie:
I finished with the PR for now (as in "I tried to implement the feature while considering all raised concerns").

I think someone should (or maybe also wanted to) test the current state, as I do not know how to test the postinstallation hook from an actual `npm install`/`yarn add`. Let me know, if the PR still needs work to be done.

HellSquirrel:
I'm still going to test it :) I can get here in a week or two

ai:
There is a simple way to test `postinstall`

1. `npm pack` to get tgz file
2. Use link to this tar file in demo project `npm install ../lefthook/lefthook.tgz`

yumauri:
Still there? :)

fleischie:
Oh hey, yes. I actually overlooked the last message from Andrey, so I will check today/tomorrow on some systems whether it works ok, and then report back. ✌️

fleischie:
I did now test my PR on mac and Windows 7 VM:
The downloading and calling of `lefthook install -f` seems to work alright. Unfortunately as my VM image is 32-bit, I couldn't verify that the complete process works.

It does for mac (if I change the `index.js` appropriately).

fleischie:
I also rebased the PR on the latest `master` branch, removed the `[WIP]` from the commit-in-question, and some further improvements to the postinstall script:
- only allow 64bit downloads on windows (there are no 32bit executables in the releases tab),
- remove the pre-build executables from the binary folder, and
- consolidate the naming of the executable (rename the downloaded executable `lefthook`/`lefthook.exe` for MacOS/Linux resp. Windows), and call them by name/platform from `index.js`.

yumauri:
I'm not sure, but is it possible to point package.json's `bin.lefthook` property right into binary, instead of `index.js`, to eliminate _node_ spawning for simply choose right binary?

fleischie:
Updated the PR again to fit snugly on the latest master.

@yumauri I am uncertain whether `package.json#bin.lefthook` can be the lefthook binary directly, as firstly the docs on that field give the impression that it must be a node script, and secondly when i attempted to do it `npm` (at least) complained about a missing script, although the binary existed. 🤷‍♂

fleischie:
Amendum: This PR works on the assumption, that binaries are always available for the most recent version as a gzipped package. That's why there is a commit aborting on 32bit Windows.

I understand that this is an added burden on the maintainer, as there are more and more diverse packages to be produced. The other solution would be to handle the zips as well, but require an additional (potentially heavy) node zip-library, which I tried to avoid until now.

Let me know what you think.

yumauri:
Github actions are publicly available ;) Theoretically it is possible to make an action to build and upload binaries _somewhere_ on every _push_

ai:
> @yumauri I am uncertain whether package.json#bin.lefthook can be the lefthook binary directly, as firstly the docs on that field give the impression that it must be a node script, and secondly when i attempted to do it npm (at least) complained about a missing script, although the binary existed. man_shrugging

You can test npm features by installing npm package from local file `"test": "../test"`

ringods:
Please, please, please do **NOT** merge this. Not everybody is working in an environment where all machines have direct internet connectivity!

I have customers in highly regulated environments where their working machines are isolated from the internet. Development is done with an internal package repository, acting as a proxy for upstream repositories. All package downloads come from this proxy.

The current situation, with the binaries included in the lefthook package, allow these customers to work with lefthook correctly. If you change it to downloading during the postinstall, their setups will break and they have no way to work around it as you hardcode the download URL.

Making the URL configurable also pushes a lot of chore onto the users as they have to copy the binaries, put them on an internal site and communicate the link. Quite error prone and not usage friendly.

To wrap up: can you please leave it as it is?

ai:
Yeap, I changed my mind. Let's close it.

privatenumber:
Seems @ringods's concern was neglected in #188:

> I have customers in highly regulated environments where their working machines are isolated from the internet. Development is done with an internal package repository, acting as a proxy for upstream repositories. All package downloads come from this proxy.

If the npm registry is proxied, I think this can be easily addressed by publishing the binaries to npm instead of using hosting them as a GitHub release.

For example, playwright does this with their browser flavors (eg. [playwright-chromium](https://www.npmjs.com/package/playwright-chromium)).

Issue #45, opened 2019-08-05
RuboCop config being ignored?


andyw8:
My RuboCop is configured to ignore schema changes:

```yml
require: 'rubocop-rails'
inherit_from: .rubocop_todo.yml

AllCops:
  TargetRubyVersion: 2.6
  Exclude:
    - 'db/schema.rb' # auto-generated
```
When I run RuboCop directly, there are no issues.

I have Lefthook configured with:

`run: bundle exec rubocop {staged_files} --parallel`

When it's triggered through Lefthook, it seems to ignore the config.

I tried changing it to explicitly specify the config file, but it made no difference:

`run: bundle exec rubocop --config .rubocop.yml {staged_files} --parallel`

Arkweid:
This is strange :thinking:  I will check it.

AquiTCD:
What about to use `--force-exclusion` option for RuboCop?
RuboCop does not ignore `Exclude` config when the command given file except to use `--force-exclusion`.

andyw8:
@AquiTCD thanks, that works.

I think others will also run into this, so I'll open a PR to document it.

PR #54, opened 2019-08-07
Add note about RuboCop configuration

andyw8:
See #45.

https://docs.rubocop.org/en/latest/configuration/#includingexcluding-files

### DIFF
--- a/docs/full_guide.md
+++ b/docs/full_guide.md
@@ -117,6 +117,8 @@ pre-commit:
       run: bundle exec rubocop {all_files} # {all_files} - list of files
 ```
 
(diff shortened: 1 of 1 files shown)


Arkweid:
@andyw8 good note. I think the best place for that note here: https://github.com/Arkweid/lefthook/blob/master/docs/full_guide.md
Because we want to keep `README.md` short and simple as possiible.
Can you move it?
Also I added your note here: https://github.com/Arkweid/lefthook/wiki/Troubleshooting

andyw8:
@Arkweid I've pushed an update.

PR #82, opened 2019-10-17
Use relative path to script/executable file

AlexeyMatskevich:
It's PR fix https://github.com/Arkweid/lefthook/issues/81
Problem: `Lefthook` launches runner passing the absolute path to the executable.
The container having the volume mounted from the host machine which provides source code has other paths to the executable files in absolute form.

Solution - use a relative path to executable files.

Perhaps a better solution would be to use a key in a config file that will be determined in what form the path should be transmitted.


Arkweid:
Great!

PR #86, opened 2019-11-08
Fixed path to binaries in postinstall script

sHooKDT:
Attempt to resolve issue https://github.com/Arkweid/lefthook/issues/85

### DIFF
--- a/.npm/postinstall.js
+++ b/.npm/postinstall.js
@@ -1,11 +1,9 @@
-require("process");
 const { spawn } = require("child_process");
 const { join } = require("path");
 
 const isCI = process.env.CI;
-if (!isCI) {
-  process.chdir(process.env.INIT_CWD);
 
+if (!isCI) {
   if (process.platform === 'darwin') {
     binary = 'lefthook-mac';
   } else if (process.platform === 'linux') {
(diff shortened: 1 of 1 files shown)


sHooKDT:
@Arkweid check this out, works from the first sight at least
I don't understand why was cwd used there, because binaries are always located next to postinstall script, so my approach should work I believe

Arkweid:
> @Arkweid check this out, works good from the first sight
> I don't understand why was cwd used there, because binaries are always located next to postinstall script, so my approach should work I believe

Cool! Will check today.

Arkweid:
Unfortenuly it doen't work :(
One underline missed
```
https://github.com/Arkweid/lefthook/pull/86/files#diff-57f92dec60c72c04bad912182b175f12R18
```
And without this line:
```
process.chdir(process.env.INIT_CWD);
```
We not return back to the root directory and postinstal script will install config and hooks itto `.node_modules` directory.

sHooKDT:
Yes, of course.

> Postinstal script will install config and hooks into .node_modules directory.

I don't completely understand what you mean here. All files inside `.npm` directory will be collocated inside `node_modules` as I know. So why do we care about configs/hooks/whatever?

sHooKDT:
I found the previous version of your comment and understood what it's about :)
So what if we just return `process.chdir(process.env.INIT_CWD);`?

PR #88, opened 2019-11-10
Fixed path to binaries

sHooKDT:
Another try @Arkweid 
I've added cwd argument to spawn, so it should work as earlier.


Arkweid:
Will check it tomorrow.
Here is a link, how to check it localy:
https://github.com/Arkweid/lefthook/pull/43#issuecomment-535441507

PR #94, opened 2019-12-01
Use eval instead of exec; Enable tty for the shell

ssnickolay:
There are minor changes to support interactive pre-commit scripts.

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -20,7 +20,7 @@ var addCmd = &cobra.Command{
 
 ├───.git
 │   └───hooks
-│       └───pre-commit // this executable will be added. Existed file with 
+│       └───pre-commit // this executable will be added. Existed file with
 │                      // same name will be renamed to pre-commit.old
 (lefthook add this dirs if you run command with -d option)
 │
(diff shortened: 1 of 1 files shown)


cmd/add.go
```diff
@@ -64,12 +64,15 @@ func addHook(hookName string, fs afero.Fs) {
...
+
+exec < /dev/tty # <- enables interactive shell
+
 ` + autoInstall(hookName, fs) + "\n" + "cmd=\"lefthook run " + hookName + " $@\"" +
 		`
 
 if lefthook >/dev/null 2>&1
 then
-  exec $cmd
+  eval $cmd
```
ssnickolay: Here is the reason of this change https://unix.stackexchange.com/a/296842

cmd/add.go
```diff
@@ -64,12 +64,15 @@ func addHook(hookName string, fs afero.Fs) {
 if [ "$LEFTHOOK" = "0" ]; then
   exit 0
 fi
+
+exec < /dev/tty # <- enables interactive shell
```
ssnickolay: This is useful when you run interactive scripts by pre-commit hook https://unix.stackexchange.com/a/91716

Issue #96, opened 2019-12-06
Behaviour of install/uninstall commands


justinasposiunas:
`install` command does not follow single responsibility principle by having a multiple behaviours depending on the state of a project:

* `install` creates a template file `lefthook.yml` in an empty project and it might be moved to `create` or `init`
* `install` creates git hooks when `lefthook.yml` is present in a project, otherwise it should say that lefthook needs to be initialized first

I came to this by noticing `uninstall` first because it removes not only files in `.git/hooks` but configuration files also and I see that as critical and unexpected behaviour.

It might be useful to have configuration files always present and having lefthook as optional tool. So everyone in a team could choose whether use it or not but keep configuration in a project.

In conclusion, I have a few proposals and I can implement any of those:

**Solution no. 1**

I would prefer this more than another one. Splitting `install` into `install` and `init` like described above also updating `uninstall` so that won't remove configuration files by default.

**Solution no. 2**

Leaving `install` as it is but updating `uninstall` by adding flags, i.e. one of those:

* `lefthook uninstall --ignore lefthook.yml lefthook-local.yml`
* `lefthook uninstall --keep-config`
* `lefthook uninstall --only-hooks`

Arkweid:
Thank you for your comment @justinasposiunas
The reason why two actions are combined here at one command (config and hooks, if there is already a config) is the postinstall action for yarn/npm package systems.
User just make `yarn add @arkweid/lefthook` and he already have initialized config.
If user download project with existed lefthook, he just make `yarn install` and because of config already exists hooks will be set.
Of course, we could put it on a postinstall wrapper, but I don't see any problems with two actions at a time (in our particular case).

So, solution no. 2 lefthook `uninstall --keep-config` looks acceptable.

justinasposiunas:
> Thank you for your comment @justinasposiunas
> The reason why two actions are combined here at one command (config and hooks, if there is already a config) is the postinstall action for yarn/npm package systems.
> User just make `yarn add @arkweid/lefthook` and he already have initialized config.
> If user download project with existed lefthook, he just make `yarn install` and because of config already exists hooks will be set.
> Of course, we could put it on a postinstall wrapper, but I don't see any problems with two actions at a time (in our particular case).
> 
> So, solution no. 2 lefthook `uninstall --keep-config` looks acceptable.

Thanks for the quick reply, got it! So I agree that 2nd option is also acceptable and will create a PR related to this soon.

Arkweid:
Exists in 0.6.7

PR #97, opened 2019-12-06
remove only hooks and keep config files present Arkweid/lefthook#96

justinasposiunas:
All the context details about the improvement is on #96


Arkweid:
Cool! I'll check it soon.

Issue #72, opened 2019-08-30
Support for relative roots/monorepos


ffloyd:
For example, when you working with monorepo you may want to have config like this:

```yaml
pre-commit:
  commands:
    elixir_format:
      root: "/backend"
      glob: "*.{ex,exs}"
      run: mix format --check-formatted {staged_files}
    ts_format:
      root: "/frontend"
      glob: "*.{ts,tsx}"
      run: npm run format_check {staged_files}
```

Possible `root` option behaviour:

* helpers like `{staged_files}` should return paths relative to provided root path
* option `files` should use root path as working dir
* option `run` should use root path as working dir

Arkweid:
#38 ref

sHooKDT:
+1 for this issue. It's a huge pain to use eslint in monorepo now.
I guess I can work on this a bit later, if nobody will take care.

Arkweid:
I give it a try in followng week

dflock:
I'd really like to start using lefthook and this is a roadblock for us.

Arkweid:
@sHooKDT makes I few changes for monorepos(big thanks), also root path in 0.6.5 now always related to git top.
Probable not a big changes we need to do to make it works:
1. Add new config key to the list (for `root: "/backend"`)
https://github.com/Arkweid/lefthook/blob/master/cmd/run.go#L42
2. Grab it from config file like this:
https://github.com/Arkweid/lefthook/blob/master/cmd/run.go#L469-L472
3. Our runners now can be executed on some context(`/backend`, `frontend`)
So we need to look at something like `chdir` https://golang.org/src/os/file.go?s=7660:7688#L251 or if you familiar with bash you can change `sh -c` i some way.
4. Double the code for `_windows.go` postfix files

jsmestad:
I used @Arkweid's guidance to implement this in #98.

* `{all_files}` and `{staged_files}` will be relative to the `root` path specified
  * For example, root `"api/"` will change "api/spec/spec_helper.rb" to "./spec/spec_helper.rb"
  * Exclude files that do not fall inside of `root` (for example, "frontend/index.js" would be excluded from the task)
* Change directory into `root` when executing command

Arkweid:
@sHooKDT @dflock @ffloyd 
It would be great if someone will also check PR for existing project

dflock:
I'll try to get to this today,

Thanks!

Arkweid:
[...]

PR #98, opened 2019-12-06
Support relative roots for monorepos (#72)

jsmestad:
I did my best as a non-golang expert to implement #72 with as little code as possible.

## Feedback

It appears there is a step missing from the `CONTRIBUTING.md` file about how to get `lefthook` into the top of `$PATH`. I had to use `gem uninstall lefthook` and then `go install` to get the specs to run. However those also failed to pass because I am assuming there is more set up to do that I must have missed.


Arkweid:
Cool! I'll check it soon.

Arkweid:
Cool! I'll check it soon.

jsmestad:
@Arkweid let me know. My team is eager to use this on our project 😉

Arkweid:
Of course. I just usually have a time on weekends.

jsmestad:
@Arkweid @dflock I found a bug with the root. It needed to be an absolute path. I have pushed the fix to this branch.

PR #103, opened 2020-01-18
Add possibility for using multiple file types (#101)

### LINKED ISSUE #101: Support `lefthook.yaml`
I already have created a `lefthook.yaml` file and when I ran `lefthook install`, it created a new `lefthook.yml` not recognizing the `.yaml` file. It would be nice to have support both extensions as both are valid and if your project already uses `.yaml`, it would be so ugly to have just one black sheep with `.yml`.

rbUUbr:
Closes #101 
Added possibility to use files for a config with `yaml` extension.
Questions to discuss:
1. I haven't added tests, because locally they are failed even on the master, Do I need to perform something special?
2. In `cmd/install.go:189` you can see `primaryMatch`, this code will just take first match fr config file. Do we need anything special for it? (IMHO, no)

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -57,7 +57,7 @@ func init() {
 
 // InstallCmdExecutor execute basic configuration
 func InstallCmdExecutor(args []string, fs afero.Fs) {
-	if yes, _ := afero.Exists(fs, getConfigYamlPath()); yes {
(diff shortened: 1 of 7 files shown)


Arkweid:
Awesome! If it not urgent I will check it soon.

PR #107, opened 2020-01-26
Add logrus integration

### LINKED ISSUE #78: Reduce the noise
There's too much noise:

```
Lefthook v0.6.3

  EXECUTE > yamllint
(SKIP. NO FILES FOR INSPECTING)

  EXECUTE > jsonnet-fmt
(SKIP. NO FILES FOR INSPECTING)

  EXECUTE > vault-policy-fmt

SUMMARY: (done in 479.47 seconds)
✔️  vault-policy-fmt
```

This is good for debugging purposes, but if you have tens of different commands and just changed a single file, seeing all this is not signal, but noise. Showing the program and the version number is unnecessary either (if one cares, they can see that with `lefthook version`).

rbUUbr:
1. Added logrus for this project
2. I've added flag for `rootCmd` with `log-level`, but seems it does not work :( could you please check it? Because I found out that other flags for `rootCmd` also doesn't work

closes  #78


Arkweid:
Awesome work. I will check it.

Arkweid:
Great idea with `loggerClient`, but I don't sure now we can use `Logrus` as a library. I went through docs and looks like `Logrus` can be configured only by levels, but we need flexible logs groups which may be combined in any positions.
https://github.com/Arkweid/lefthook/issues/78#issuecomment-538277400

rbUUbr:
hm, interesting, I thought, that we can different log levels as log groups. for example, the summary will have an info log level, failures will have an error log level. Maybe you're right, but as for me (as for active user of Lefthook), it is not too interesting to output just failures, but interesting to get some general information with failures. Can we do this? or how do you see this feature at all?

Arkweid:
I see it like this:
```go
// somewhere in the code:
loggerClient.Meta(au.Cyan("SYNCING"), au.Bold("lefthook.yml"))

// in logger client we checking config key `output`
if outputMetaEnabled {
        log.Println(args...)
}
```
For failure/success keys we store output from processes in pseudo terminal and write in after command ended with some exit code

rbUUbr:
yep, that's it! for info log level you will receive errors also, but if you will set log level to error or to warn, you will not get meta information output, that's the way how we can combine outputs. It is imo, we can do it through yml config like here https://github.com/Arkweid/lefthook/issues/78#issuecomment-538277400

Arkweid:
If you sure we can made it with logrus ok. Could you make example in this PR?

rbUUbr:
hm, at first, I think, it will be very good to solve issues that I've provided in general comment, because they're some kind of blockers for this :D 
> I've added flag for rootCmd with log-level, but seems it does not work :( could you please check it? Because I found out that other flags for rootCmd also doesn't work

could you please check it?

Arkweid:
> could you please check it?
Ok! I will check it. Just a little busy during the workdays :sweat_smile:

Arkweid:
Sorry for so long @rbUUbr !
I have cheked it and logrus not fit well our desires. For this task we don't need `log-level`.
We only just need to mark every output with their purpose in bloks. Like it described here:
https://github.com/Arkweid/lefthook/issues/78#issuecomment-538277400
For example:
```
loggerClient.Meta(au.Cyan("Lefthook v" + version))

loggerClient.Summary(au.Cyan("\nSUMMARY:"), au.Brown("(SKIP EMPTY)"))
```

Arkweid:
Solved.

PR #108, opened 2020-01-26
Fix execution on windows

### LINKED ISSUE #90: `lefthook version` fails with `exit status 128` outside of git repo
Hello,

Context: 
* Windows 10
* Build 0.6.5 downloaded from github `lefthook_0.6.5_Windows_x86_64.zip`

Problem: `lefthook.exe version` should return the current version. 

### Actual behavior
In a non git repo. without `lefthook.yml`, `exit status 128` is returned 
In a git repo. with  `lefthook.yml`, `0.6.5` is returned.

### Waited behavior
`0.6.5` must be returned 

Herve-M

### LINKED ISSUE #91: sh executable file not found
On Windows 10 with lefthook 0.6.5, I get the following error:

```
exec: "sh": executable file not found in %PATH%
```

It even occurs when running `lefthook version`.

lionskape:
removed sh from dependencies in root and add commands.

fixes #90
fixes #91

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -140,8 +140,7 @@ func addLocalHookDir(hookName string, fs afero.Fs) {
 }
 
 func getGitHooksDir() string {
-	commandArg := strings.Split("git rev-parse --git-common-dir", " ")
-	cmd := exec.Command(commandArg[0], commandArg[1:]...)
+	cmd := exec.Command("git", "rev-parse", "--git-common-dir")
 
 	outputBytes, err := cmd.CombinedOutput()
 	if err != nil {
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -140,7 +140,8 @@ func addLocalHookDir(hookName string, fs afero.Fs) {
 }
 
 func getGitHooksDir() string {
-	cmd := exec.Command("sh", "-c", "git rev-parse --git-common-dir")
+	commandArg := strings.Split("git rev-parse --git-common-dir", " ")
+	cmd := exec.Command(commandArg[0], commandArg[1:]...)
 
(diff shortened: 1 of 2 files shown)


cmd/add.go
```diff
@@ -140,7 +140,8 @@ func addLocalHookDir(hookName string, fs afero.Fs) {
 }
 
 func getGitHooksDir() string {
-	cmd := exec.Command("sh", "-c", "git rev-parse --git-common-dir")
+	commandArg := strings.Split("git rev-parse --git-common-dir", " ")
+	cmd := exec.Command(commandArg[0], commandArg[1:]...)
```
Arkweid: We have static command here, could you please change it in this way:
```go
cmd := exec.Command("git", "rev-parse", "--git-common-dir")
```

cmd/root.go
```diff
@@ -109,7 +109,8 @@ func getRootPath() string {
 
 func setRootPath(path string) {
 	// get absolute path to .git dir (project root)
-	cmd := exec.Command("sh", "-c", "git rev-parse --show-toplevel")
+	commandArg := strings.Split("git rev-parse --show-toplevel", " ")
+	cmd := exec.Command(commandArg[0], commandArg[1:]...)
```
Arkweid: We have static command here, could you please change it in this way:
```go
cmd := exec.Command("git", "rev-parse", "--show-toplevel")
```

cmd/add.go
lionskape: Fixed

cmd/root.go
lionskape: fixed

PR #118, opened 2020-03-09
WIP: Add support for remote config

zephraph:
👋 

I've been investigating adding lefthook to @artsy's repos, but thought it'd be nice to have some centralized config. I see you already support `extends`, but that relies on the files being already present. 

This PR will (hopefully) add the ability to specify `extends` as a URL. 

```yaml
extends: https://raw.githubusercontent.com/artsy/.github/master/base.yml
```

This is my first time ever using Go so please excuse any obvious blunders 🙏. I'll gladly update anything that doesn't match expectations/style of the project.

I still consider this a bit of a work-in-progress but I'd appreciate some early feedback. 

## Todo

- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -165,13 +165,10 @@ func EnableColors() bool {
 
 func fetchRemoteExtends(uri string) io.Reader {
 	resp, err := http.Get(uri)
-	defer resp.Body.Close()
 	if err != nil {
 		log.Fatal("Error fetching remote config", uri, "\n", err)
 	}
-	if err != nil {
-		log.Fatal("Error reading remote config", uri, "\n", err)
-	}
+	defer resp.Body.Close()
 	return resp.Body
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"os/exec"
 	"io"
+	"io/ioutil"
+	"bytes"
 	"path/filepath"
 	"strings"
 	"net/http"
@@ -169,7 +171,11 @@ func fetchRemoteExtends(uri string) io.Reader {
 		log.Fatal("Error fetching remote config", uri, "\n", err)
 	}
 	defer resp.Body.Close()
-	return resp.Body
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatal("Error reading remote config", uri, "\n", err)
+	}
+	return bytes.NewReader(body)
 }
 
 func isValidUrl(maybeUrl string) bool {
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -93,13 +96,20 @@ func initConfig() {
...
-		extension := filepath.Ext(getExtendsPath())
-		name := filename[0 : len(filename)-len(extension)]
-		viper.SetConfigName(name)
-		viper.AddConfigPath(filepath.Dir(getExtendsPath()))
-		err := viper.MergeInConfig()
-		check(err)
+		if isValidUrl(getExtendsPath()) {
+			uri := getExtendsPath()
+			baseRemoteConfig := fetchRemoteExtends(uri)
+			err := viper.MergeConfig(baseRemoteConfig)
```
zephraph: I didn't call `viper.SetConfigName` prior to this. I'm not sure if that's necessary or not.

Arkweid:
Hi @zephraph 
Right now Lefthook sync changes between config and hooks dir automaticly with every run. How do you plan to handle this for a remote config?

zephraph:
I’m not sure 100% understand the question. The current implementation for the remote config is a bit naive. It just downloads the config on every run. I’m sure we’d want to ultimately implement a mechanism to pull down the remote config as save it locally to be used later. Figuring out when to update the local copy is a good question. 

I don’t have a lot of solid opinions on how to handle that. I’m more than happy to work towards a solution that you’d like to see though.

cmd/root.go
```diff
@@ -93,13 +98,21 @@ func initConfig() {
...
-		name := filename[0 : len(filename)-len(extension)]
-		viper.SetConfigName(name)
-		viper.AddConfigPath(filepath.Dir(getExtendsPath()))
-		err := viper.MergeInConfig()
-		check(err)
+		if isValidUrl(getExtendsPath()) {
+			uri := getExtendsPath()
+			baseRemoteConfig := fetchRemoteExtends(uri)
+			viper.SetConfigType("yaml")
+			err := viper.MergeConfig(baseRemoteConfig)
```
Arkweid: Just merge is not enough here. We need to install that config. Thus way we'll have local changes in `.git/hooks`

Arkweid: How local config will interact with remote?

zephraph: > Just merge is not enough here. We need to install that config. Thus way we'll have local changes in .git/hooks

I don't exactly know the answer to this. I assumed that given it's pulled in an merged when the command is ran that whenever the regular config is installed this would be too. I'll have to take some more time to really understand how the config is installed to definitively answer that question.

zephraph:
I don't really have the time to finish this, so I'm gonna close.

PR #126, opened 2020-05-08
Feature: Ability to use multiple extends

Evilweed:
Hello,
I would like to add the ability to use multiple extends. This was requested as a feature at my company as we are using Monorepo with many different projects (that are not separate git submodules), and using just one YML to manage all of those projects makes it become some GOD huge and unreadable file.

**With this PR in effect, you can define and use multiple extends like this:**
```
extends:
  - landing/lefthook-landing.yml
  - backend/lefthook-backend.yml
```

@Arkweid 🙏

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -100,7 +100,7 @@ func TestExtendsProperty(t *testing.T) {
 	viper.SetConfigType("yaml")
 
 	viper.ReadConfig(bytes.NewBuffer([]byte("")))
-	assert.True(t, isConfigExtends(), "Should not detect extends property")
+	assert.False(t, isConfigExtends(), "Should not detect extends property")
 
 	viper.ReadConfig(bytes.NewBuffer(yamlExampleString))
 	paths := getExtendsPath()
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -93,13 +93,15 @@ func initConfig() {
 	viper.MergeInConfig()
 
 	if isConfigExtends() {
-		filename := filepath.Base(getExtendsPath())
-		extension := filepath.Ext(getExtendsPath())
-		name := filename[0 : len(filename)-len(extension)]
-		viper.SetConfigName(name)
-		viper.AddConfigPath(filepath.Dir(getExtendsPath()))
-		err := viper.MergeInConfig()
-		check(err)
+		for _, v := range getExtendsPath() {
+			filename := filepath.Base(v)
+			extension := filepath.Ext(v)
+			name := filename[0 : len(filename)-len(extension)]
+			viper.SetConfigName(name)
+			viper.AddConfigPath(filepath.Dir(v))
+			err := viper.MergeInConfig()
+			check(err)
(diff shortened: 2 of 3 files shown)


Arkweid:
Thx for PR @Evilweed !
I will check it one current week, and if all ok going to made new release.

Arkweid:
@Evilweed looks fine! Could you add any tests? :)

Evilweed:
@Arkweid 
1. Tests are added;
2. It was also tested manually, and it is backward compatible - you can either add 1 extend or multiple in array :)

Arkweid:
> @Arkweid
> 
> 1. Tests are added;
> 2. It was also tested manually, and it is backward compatible - you can either add 1 extend or multiple in array :)

Gj! Merged.

molszanski:
Yay!

D1no:
Just to clarify: This is not an extend but an inherit (left join), correct? When we have one root lefthook.yml that extends both a node js server and client project, they will overwrite one another and not both execute? I.e. when the command "format" is used in both like so:

```yml
# client/lefthook.yml
pre-commit:
  commands:
    format: # <-----------
      glob: "*.{js}"
      run: npm ....
```

```yml
# server/lefthook.yml
pre-commit:
  commands:
    format: # <-----------
      glob: "*.{ts}"
      run: yarn ....
```

```yml
# lefthook.yml
extends:
  - client/lefthook.yml
  - server/lefthook.yml
```

What will happen in this case to `format`? Do I need to tell engineers to make sure that commands are globally unique in the mono repo? Would be great if this is a true extend also with the ability to scope the ymls globs once to folders.

_Couldn't find anything in the documentation — just evaluating lefthook right now and stumbled over this PR. For now I wrote this in an evaluation project:_

```yml
# Workspace Specific Git Hooks (using lefthook)
#
#   This file is registered in:
#   infrastructure/development/githooks.yml
#
# Warning about Adding Hooks
#
#   (1) Make sure to use the glob pattern to limit the githook to this
#       workspace only. Do "platform/browser/**/*.{js,ts,...}" not "*.{js..".
#       Otherwise the hook applies and conflicts outside this workspace folder.
#   (2) Use unique command names to avoid overwriting other githooks from other
#       workspaces with the same name. I.e. "platform-browser-prettier" instead
#       of just "prettier", which would be also used by a node.js server folder.
#

pre-commit:
...
```

mrexox:
@D1no , you are right, this is more like a merge respecting the fields from the extends. The latter extend will overwrite the previous extends in your example. It's up to user to control this behavior.

I guess that you'd expect your example to finally show something like this?

```yml
# result
pre-commit:
  commands:
    format: 
      - glob: "*.{js}"
        run: npm ....
      - glob: "*.{ts}"
        run: yarn ....
```

If you feel the need for that please create a separate feature request. And I'll add it to my radar if many people want it because it feels like a big change in the lefthook design.

By the way, if you want to test what you get after all extends use `lefthook dump` command – it will print the config file after applying all extends.

Issue #135, opened 2020-07-09
"extends" path using an environment variable not working as expected


qjcg:
I'm trying to get lefthook to use hooks from a custom external directory defined via an environment variable
(e.g. `extends: $HOOKDIR/hooks.yml`).

So far I haven't been able to get it working, but maybe I'm missing something obvious. 

Steps to reproduce (tested using bash):

```sh
# 1. A basic hook for testing purposes.
$ cat hooks.yml 
pre-commit:
  commands:
    greet:
      run: echo 'Extended hooks!'

# 2. Export a HOOKDIR environment variable and copy our hook file there.
$ export HOOKDIR=/tmp
$ cp hooks.yml $HOOKDIR

# 3. Tell lefthook about our external hooks file and test running the hook. It fails :(
$ echo 'extends: $HOOKDIR/hooks.yml' > lefthook.yml
$ lefthook run pre-commit
runtime error: slice bounds out of range [:-1]

# 4. Repeat steps 2 & 3 using $HOME instead of $HOOKDIR. It works! :)
$ cp hooks.yml $HOME
$ echo 'extends: $HOME/hooks.yml' > lefthook.yml
$ lefthook run pre-commit
Lefthook v0.7.2
RUNNING HOOKS GROUP: pre-commit

  EXECUTE > greet
Extended hooks!

SUMMARY: (done in 0.02 seconds)
✔️  greet
```

qjcg:
Looking into this, the **root cause of this issue is `viper`'s [`absPathify`](https://github.com/spf13/viper/blob/eabbc68a3ecd5cf8c11a2f84dbda5e7a38493b2f/util.go#L91) function** (as called by the [`AddConfigPath`](https://github.com/spf13/viper/blob/13494e8047bb3f4c39087dfce593dfc3a5d02260/viper.go#L465) method). It currently fails when provided with a path containing an environment variable with a single path separator _pre-expansion_ (excluding `$HOME`, which is [handled differently from other env vars](https://github.com/spf13/viper/blob/eabbc68a3ecd5cf8c11a2f84dbda5e7a38493b2f/util.go#L94)). 

So:
- `$HOOKDIR/hooks.yml` will **fail** (one pre-expansion path separator), but
- `$HOOKDIR/shared/hooks.yml` (2 pre-expansion path separators) will **succeed**

[...]

PR #141, opened 2020-07-31
Add note about using extends with path expansion

qjcg:
This documents the issue and workaround described here: https://github.com/Arkweid/lefthook/issues/135#issuecomment-667232982 .


qjcg:
Update: Finally, I got a fix to `viper` merged upstream (https://github.com/spf13/viper/pull/940), so this PR is no longer needed 🎉 

Fixes #135.

PR #142, opened 2020-08-02
feat(commands): added Git hook argument handling

### LINKED ISSUE #133: Can't retrieve arguments
I need to create a `post-checkout` hook to compare the old `HEAD` to the new one.

I've installed lefthook through `brew install Arkweid/lefthook/lefthook` (v0.7.2) and created this configuration:

```
post-checkout:
  commands:
    test:
        run: echo $1 $2 $3
```

I expect the positional arguments to contain [the values provided by the hook](https://git-scm.com/docs/githooks#_post_checkout):

> The hook is given three parameters: the ref of the previous HEAD, the ref of the new HEAD (which may or may not have changed), and a flag indicating whether the checkout was a branch checkout (changing branches, flag=1) or a file checkout (retrieving a file from the index, flag=0).

However, when running `lefthook run post-checkout val1 val2 val3`, the output is empty.

I'm not sure if this is a bug or if I'm doing something wrong?

pregnor:
**Type**: feature
**Scope**: commands
**Related issues**: fixes #133

### What is this?
Added functionality to propagate Git hook script invocation arguments to Lefthook commands in the form of positional shorthands such as `{1}` for the first argument, `{2}` for the second, etc.
Also introduced `{0}` as the equivalent shorthand of `{1} {2} ...` (space joint string of individual Git hook arguments) so all arguments can be passed down to command calls with one shorthand regardless of the actual argument count.

### Why?
To improve script vs. command feature parity and make commands able to use Git hook arguments directly.

Some Git hooks pass down arguments to the hook scripts to specify what is happening exactly.
An example is `commit-msg` where the file containing the proposed commit message is passed down as a positional argument - because it can be `.git/COMMIT_EDITMSG`, `.git/GITGUI_MSG` or another source.
In such cases this argument is best to be passed transparently to commands by Lefthook as well in order to be able to define simple commands using the exact proposed commit message or its file from the argument to do their work.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -156,9 +156,9 @@ func RunCmdExecutor(args []string, fs afero.Fs) error {
 		for _, commandName := range commands {
 			wg.Add(1)
 			if getParallel(hooksGroup) {
-				go executeCommand(hooksGroup, commandName, &wg)
+				go executeCommand(hooksGroup, commandName, &wg, gitArgs)
 			} else {
-				executeCommand(hooksGroup, commandName, &wg)
+				executeCommand(hooksGroup, commandName, &wg, gitArgs)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


Arkweid:
Awesome! Will look at soon.

PR #154, opened 2020-10-13
Fix incompatible Bash features used in SH scripts

nesk:
This PR fixes bugs introduced in https://github.com/Arkweid/lefthook/pull/146#issuecomment-695076730. 

[The SH script used for NPM](https://github.com/Arkweid/lefthook/blob/58f641c74298adf7fb7fd7ae117875271aaa4c63/.npm/bin/lefthook) was written with poor knowledges of the differences between Bash and SH. I did some research and used https://www.shellcheck.net/ to help me rewrite properly the script and stay POSIX-compliant.

Please, tell me if you see any more issues.


Arkweid:
Awesome! I'll test it more a little later.

PR #171, opened 2021-04-26
Improve check for installed git

### LINKED ISSUE #90: `lefthook version` fails with `exit status 128` outside of git repo

### LINKED ISSUE #114: Running on Docker returning status 128
Try to install on docker is giving

```
Step 15/17 : RUN lefthook install -f
 ---> Running in 945f8062d6ce
/usr/src/gems/ruby/2.7.0/gems/lefthook-0.7.1/libexec/lefthook-linux install -f
exit status 128
ERROR: Service 'web' failed to build: The command '/bin/sh -c lefthook install -f' returned a non-zero code: 1
```

DmitryTsepelev:
Closes #90, closes #114.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/logrusorgru/aurora"
+	"github.com/spf13/afero"
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -43,6 +44,21 @@ var rootCmd = &cobra.Command{
 	Long: `After installation go to your project directory
 and execute the following command:
 lefthook install`,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		if cmd.Name() == "help" || cmd.Name() == "version" {
+			return
+		}
+
+		var fs = afero.NewOsFs()
+		if gitInitialized, _ := afero.DirExists(fs, filepath.Join(getRootPath(), ".git")); gitInitialized {
+			return
+		}
+
+		message := `This command must be executed within git repository.
+Change working directory or initialize new repository with 'git init'.`
+
+			log.Fatal(au.Brown(message))
+	},
 }
 
 func Execute() {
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -43,6 +44,13 @@ var rootCmd = &cobra.Command{
 	Long: `After installation go to your project directory
 and execute the following command:
 lefthook install`,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		var fs = afero.NewOsFs()
+		gitInitialized, _ := afero.DirExists(fs, filepath.Join(getRootPath(), ".git"))
+		if !gitInitialized {
+			log.Fatal("git is not initialized")
```
Envek, line 51: That message should be more verbose to give user a hint what to do. What `git is not initialized` does mean, after all?
Maybe something like that?
```suggestion
			log.Fatal("This command must be executed within git repository. Change working directory or initialize new repository with `git init`.")
```

Envek (changes requested):
Also `lefthook help` should be fixed also, however it looks like it is a built-in command.

cmd/root.go
```diff
@@ -43,6 +44,21 @@ var rootCmd = &cobra.Command{
 	Long: `After installation go to your project directory
 and execute the following command:
 lefthook install`,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		if cmd.Name() == "help" || cmd.Name() == "version" {
```
DmitryTsepelev, line 48: this is a bit brittle but couldn't find a better way

cmd/root.go
```diff
@@ -115,8 +131,7 @@ func setRootPath(path string) {
 	// get absolute path to .git dir (project root)
 	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
 
-	outputBytes, err := cmd.CombinedOutput()
-	check(err)
+	outputBytes, _ := cmd.CombinedOutput()
```
DmitryTsepelev, line 134: we don't check the error because we either checked it in the global `PersistentPreRun` or don't care about git (i.e., in `help` or `version`)

cmd/root.go
DmitryTsepelev: ✅

DmitryTsepelev:
Yeah it was a Work in progress PR, I guess it's ready now

skryukov (approved):
lgtm

Envek (approved):

Issue #102, opened 2020-01-10
Rebasing is slow and noisy


nc-holodakg:
When rebasing, lefthook appears to be re-installing itself on every commit. I see a lot of console output that looks like this:
```
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac run prepare-commit-msg .git/COMMIT_EDITMSG message
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac run prepare-commit-msg .git/COMMIT_EDITMSG message
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac run prepare-commit-msg .git/COMMIT_EDITMSG message
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac run prepare-commit-msg .git/COMMIT_EDITMSG message
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac run prepare-commit-msg .git/COMMIT_EDITMSG message
/Users/holodakg/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/lefthook-0.6.7/libexec/lefthook-mac install
[...]

PR #173, opened 2021-04-28
Allow skipping hooks in certain git states: merge and/or rebase

### LINKED ISSUE #115: Is there a way to not run pre-commit on merge?
Hello, is there a nice way to setup Lefthook in a way that makes it not run pre-commit hooks on merge commits?

DmitryTsepelev:
`skip` now allows to skip a certain hook during merge or rebase, configuration example:

```
pre-commit:
  commands:
    commit-colors:
      run: commit-colors $(git rev-parse HEAD)
  skip:
    - merge
    - rebase
```

Things to do:

- [x] make it work on Windows
- [x] update docs

Things to discuss:

- should we refactor run and run_windows to avoid duplicating the code?
- ~is there a better name than `skip_git_states`?~ using old `skip` option
- is there a better way to determine that merge or rebase are happening?

Closes #115

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -409,12 +409,15 @@ func isScriptExist(hooksGroup, executableName string) bool {
 
 func isSkipScript(hooksGroup, executableName string) bool {
 	key := strings.Join([]string{hooksGroup, scriptsConfigKey, executableName, skipConfigKey}, ".")
-	return viper.GetBool(key)
+	return isSkip(key)
 }
 
 func isSkipCommand(hooksGroup, executableName string) bool {
 	key := strings.Join([]string{hooksGroup, commandsConfigKey, executableName, skipConfigKey}, ".")
+	return isSkip(key)
+}
 
+func isSkip(key string) bool {
 	value := viper.Get(key)
 
 	switch typedValue := value.(type) {
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -458,7 +458,7 @@ func isSkippedGitState(state string) bool {
 }
 
 func isMergeInProgress() bool {
-	if _, err := os.Stat(filepath.Join(getRootPath(), ".git", "MERGE_HEAD")); os.IsNotExist(err) {
+	if _, err := os.Stat(filepath.Join(getRootPath(), getGitDir(), "MERGE_HEAD")); os.IsNotExist(err) {
 		return false
 	}
 
@@ -466,8 +466,8 @@ func isMergeInProgress() bool {
 }
 
 func isRebaseInProgress() bool {
-	if _, mergeErr := os.Stat(filepath.Join(getRootPath(), ".git", "rebase-merge")); os.IsNotExist(mergeErr) {
-		if _, applyErr := os.Stat(filepath.Join(getRootPath(), ".git", "rebase-apply")); os.IsNotExist(applyErr) {
+	if _, mergeErr := os.Stat(filepath.Join(getRootPath(), getGitDir(), "rebase-merge")); os.IsNotExist(mergeErr) {
+		if _, applyErr := os.Stat(filepath.Join(getRootPath(), getGitDir(), "rebase-apply")); os.IsNotExist(applyErr) {
 			return false
 		}
 	}
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -458,7 +458,7 @@ func isSkippedGitState(state string) bool {
 }
 
 func isMergeInProgress() bool {
-	if _, err := os.Stat(filepath.Join(getRootPath(), getGitDir(), "MERGE_HEAD")); os.IsNotExist(err) {
+	if _, err := os.Stat(filepath.Join(getGitDir(), "MERGE_HEAD")); os.IsNotExist(err) {
 		return false
 	}
 
@@ -466,8 +466,8 @@ func isMergeInProgress() bool {
 }
 
 func isRebaseInProgress() bool {
-	if _, mergeErr := os.Stat(filepath.Join(getRootPath(), getGitDir(), "rebase-merge")); os.IsNotExist(mergeErr) {
-		if _, applyErr := os.Stat(filepath.Join(getRootPath(), getGitDir(), "rebase-apply")); os.IsNotExist(applyErr) {
+	if _, mergeErr := os.Stat(filepath.Join(getGitDir(), "rebase-merge")); os.IsNotExist(mergeErr) {
+		if _, applyErr := os.Stat(filepath.Join(getGitDir(), "rebase-apply")); os.IsNotExist(applyErr) {
 			return false
 		}
 	}
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -414,7 +414,65 @@ func isSkipScript(hooksGroup, executableName string) bool {
 
 func isSkipCommand(hooksGroup, executableName string) bool {
 	key := strings.Join([]string{hooksGroup, commandsConfigKey, executableName, skipConfigKey}, ".")
-	return viper.GetBool(key)
+
+	value := viper.Get(key)
+
+	switch typedValue := value.(type) {
+	case bool:
+		/*
+			pre-push:
+				commands:
+					packages-audit:
+						skip: true
+		*/
+		return typedValue
+	case string:
+		/*
+			pre-push:
+				commands:
+					packages-audit:
+						skip: merge
+		*/
+		return isSkippedGitState(typedValue)
+	case []interface{}:
+		/*
+			pre-push:
+				commands:
+					packages-audit:
+						skip:
+							- merge
+							- rebase
+		*/
+		for _, gitState := range typedValue {
+			if isSkippedGitState(gitState.(string)) {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func isSkippedGitState(state string) bool {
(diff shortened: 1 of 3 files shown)


Envek:
Oh, yes, I want it so bad!

This is especially useful for `post-checkout` hook (as in our [Lefthook, Crystalball, and git magic for smooth development experience](https://evilmartians.com/chronicles/lefthook-crystalball-and-git-magic) blog post) as usually  you don't want to run such things during `git pull --rebase` and so on.

And this is a good reason to bump version :-)

---

> is there a better name than `skip_git_states`?

What about using existing boolean `skip` directive (allow it to also receive strings or arrays)? Something like this:
```yaml
pre-push:
  commands:
    packages-audit:
      skip:
       - merge
       - rebase
```

DmitryTsepelev:
Changes:

- renamed the option to `skip`;
- added docs;
- moved duplicated functions from `run` and run_windows to a separate file (not sure what I'm doing).

Announcement: I have never tested it on windows!

aminya:
Also fixes #102

cmd/run_helpers.go
```diff
@@ -0,0 +1,340 @@
...
+
+	return false
+}
+
+func isSkippedGitState(state string) bool {
+	return state == "merge" && isMergeInProgress() || state == "rebase" && isRebaseInProgress()
+}
+
+func isMergeInProgress() bool {
+	if _, err := os.Stat(".git/MERGE_HEAD"); os.IsNotExist(err) {
```
skryukov, line 62: We should use `getRootPath()`

cmd/run_helpers.go
```diff
@@ -0,0 +1,340 @@
...
+		return false
+	}
+
+	return true
+}
+
+func isRebaseInProgress() bool {
+	if _, mergeErr := os.Stat(".git/rebase-merge"); os.IsNotExist(mergeErr) {
```
skryukov, line 70: Here is another way of checking rebasing state (not tested personally, though):
https://github.com/istio/old_mixer_repo/pull/262/files#diff-7ee9e529af99bc938983416b1810255ecd127a0f928d5c7c9d9c9d8359593d3fR28

Envek: Detached HEAD state may occur in many ways. User can manually go to it executing `git checkout somecommitsha`, so no, we shouldn't rely on it.

Envek (changes requested):
Build with `goreleaser build --skip-validate --rm-dist` isn't passing with following error:

```
   ⨯ build failed after 1.67s error=failed to build for windows_amd64: # github.com/evilmartians/lefthook/cmd
cmd/run_windows.go:14:2: imported and not used: "sort"
cmd/run_windows.go:22:2: imported and not used: "github.com/gobwas/glob"
```

cmd/run_helpers.go
DmitryTsepelev: @Envek what do you think of my approach?

DmitryTsepelev:
Had to make a new branch and copy changes since there were lots of conflicts 🙈

Envek (changes requested):
Seems like there is some validation of config types exists that prevents it from working.

Given following `lefthook.yml`:
```yml
post-checkout:
  piped: true
  scripts:
    01-bundle-checkinstall:
      skip: true
      tags: backend
    02-db-migrate:
      skip: true
      tags: backend
    03-crystalball-update:
      skip: true
      tags: rspec backend
```

And following `lefthook-local.yml`:
```yml
post-checkout:
  piped: true
  scripts:
    01-bundle-checkinstall:
      skip: false
    02-db-migrate:
      skip: rebase
    03-crystalball-update:
      skip:
        - rebase
        - merge
```

Checkout to another branch (without merge or rebase in progress) produces following output:

```
$ git checkout master

ERROR 2021/05/22 12:16:28 svType != tvType; key=skip, st=string, tt=bool, sv=rebase, tv=true
ERROR 2021/05/22 12:16:28 svType != tvType; key=skip, st=[]interface {}, tt=bool, sv=[rebase merge], tv=true
Lefthook v0.7.5
RUNNING HOOKS GROUP: post-checkout

  EXECUTE > 01-bundle-checkinstall
The Gemfile's dependencies are satisfied

02-db-migrate (SKIP BY SETTINGS)

03-crystalball-update (SKIP BY SETTINGS)

SUMMARY: (done in 0.36 seconds)
```

**Expected result**: all steps are run

**Actual result**: steps 2 and 3 are skipped, type errors are printed to console

cmd/run.go
```diff
@@ -414,7 +414,65 @@ func isSkipScript(hooksGroup, executableName string) bool {
...
+
+	return false
+}
+
+func isSkippedGitState(state string) bool {
+	return state == "merge" && isMergeInProgress() || state == "rebase" && isRebaseInProgress()
+}
+
+func isMergeInProgress() bool {
+	if _, err := os.Stat(filepath.Join(getRootPath(), ".git", "MERGE_HEAD")); os.IsNotExist(err) {
```
Envek, line 461: Inside git worktrees, `getRootPath(), ".git"` will be a file, not directory! Also every worktree has its own git directory placed _inside_ `.git` directory of the main worktree :exploding_head:

cmd/run.go
```diff
@@ -414,7 +414,65 @@ func isSkipScript(hooksGroup, executableName string) bool {
...
+		return false
+	}
+
+	return true
+}
+
+func isRebaseInProgress() bool {
+	if _, mergeErr := os.Stat(filepath.Join(getRootPath(), ".git", "rebase-merge")); os.IsNotExist(mergeErr) {
+		if _, applyErr := os.Stat(filepath.Join(getRootPath(), ".git", "rebase-apply")); os.IsNotExist(applyErr) {
```
Envek, line 470: I wonder, whether can we check for single `REBASE_HEAD` file instead of this two checks? :thinking:

Also here we need to use output of `git rev-parse --git-dir` command to get `.git` directory for worktrees.

Envek (changes requested):
One more thing is git worktree support.

As every worktree can do rebases and merges independently from main worktree, every worktree has its own git directory (which can be retrieved with `git rev-parse --git-dir` command) where we need to search for merge and rebase files (however, hooks are global for all worktrees and are only stored in global worktree).

For example, if we have repository `~/foo` with worktree `~/foo-wt` and both are during rebase, directory structure will be as follows:

 1. For main repo:
    - root url: `~/foo`
    - git dir: `~/foo/.git` (here will be `REBASE_HEAD` for main tree)
    - git hooks dir: `~/foo/.git/hooks`
 2. For worktree `foo-wt`:
    - root url: `~/foo-wt`
    - git dir: `~/foo/.git/worktrees/foo-wt` (here will be `REBASE_HEAD` for worktree)
    - git hooks dir: `~/foo/.git/hooks` (same as for the main repo)

These are output of `git rev-parse --show-toplevel`, `git rev-parse --git-dir`, and `git rev-parse --git-path hooks` commands respectively.

Looks like that we need one more helper like `getGitDir` that will return path from `git rev-parse --git-dir` command (see https://github.com/evilmartians/lefthook/pull/177 for reference).

cmd/run_helpers.go
Envek: I think that it is totally right. Not sure about details: should we check for these `rebase-merge` and `rebase-apply` directories or for `REBASE_HEAD` file.

cmd/run.go
DmitryTsepelev: [Looks like](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---show-current-patch) REBASE_HEAD exists only when it's an "interactive rebase or when rebase is stopped because of conflicts"

DmitryTsepelev:
Talkin about merging configs: looks like Viper [does not support](https://github.com/spf13/viper/blob/cdb5e5976f9bb9e9ede7da170187742fc4bd7e9f/viper.go#L1829) merging values of different types. Looks like we have to either change our API (e.g., bring back a separate `skip_git_stages` key) or hack around Viper (because I believe they won't accept a PR allowing to prefer boolean over anything else). Thoughts?

cmd/run.go
DmitryTsepelev: Fixed

Envek:
For now I created issue in Viper for this (https://github.com/spf13/viper/issues/1142) as I wish that we could reuse the same `skip` setting.

However, it is probably quite a rare use case (when hooks are always disabled in main config, and only during merge/rebase in local config), so may be we can safely ignore it for now?

DmitryTsepelev:
Yeah, we can add a separate issue for that and see if it's a big deal. Would be really nice for Viper to handle it for us!

cmd/run.go
Envek: Yeah, official git prompts also checks for these directories, so let's keep it as it is: https://github.com/git/git/blob/de88ac70f3a801262eb3aa087e5d9a712be0a54a/contrib/completion/git-prompt.sh#L446-L462

cmd/run.go
Envek: And fixed one more time for worktrees (pushed commit right into your pull request, sorry!)

Envek (approved):

Envek:
Thank you, @DmitryTsepelev, amazing work!

Envek:
Finally released in [0.8.0](https://github.com/evilmartians/lefthook/releases/tag/v0.8.0)

PR #175, opened 2021-05-02
Fix RSpec

skryukov:
This PR fixes the broken RSpec suite. We'd port it to the native go tests later. Or can we consider removing it instead of fixing it?

### DIFF
--- a/spec/support/file_structure.rb
+++ b/spec/support/file_structure.rb
@@ -8,7 +8,7 @@ class FileStructure
 
     def have_git
       FileUtils.mkdir_p(File.join(tmp))
-      _, _, _ = Open3.capture3('git init', chdir: FileStructure.tmp)
+      _, _, _ = Open3.capture3('git init', chdir: tmp)
     end
 
     def make_scripts_preset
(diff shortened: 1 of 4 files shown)


DmitryTsepelev (approved):
I'd prefer to keep both suites for now and port—at least we have more checks to make things harder to break

spec/support/file_structure.rb
```diff
@@ -7,7 +7,8 @@ class << self
     attr_accessor :root
 
     def have_git
-      FileUtils.mkdir_p(File.join(tmp, '.git', 'hooks'))
+      FileUtils.mkdir_p(File.join(tmp))
+      _, _, _ = Open3.capture3('git init', chdir: FileStructure.tmp)
```
Envek, line 11: Is there any reason to reference `tmp` method with implicit and explicit receiver in adjacent lines? It is pretty confusing.
```suggestion
      FileUtils.mkdir_p(File.join(tmp))
      _, _, _ = Open3.capture3('git init', chdir: tmp)
```

Envek (approved):

PR #176, opened 2021-05-03
`getGitHooksPath` helper function

### LINKED ISSUE #170: Custom hooks path isn't supported
From [githooks documentation](https://git-scm.com/docs/githooks):
> By default the hooks directory is `$GIT_DIR/hooks`, but that can be changed via the `core.hooksPath` configuration variable (see [git-config[1]](https://git-scm.com/docs/git-config)).

I'm not sure about use cases for this and most probably if someone moves hooks directory to another place, they don't need Lefthook at all (cause they just can commit raw hook scripts right into the repo).

But, I can see in our codebase that in various places git hooks directory detection is done very different, e.g.:
https://github.com/evilmartians/lefthook/blob/04f9b560add658c45f226c2d524d7040d4b405c2/cmd/add.go#L154
https://github.com/evilmartians/lefthook/blob/04f9b560add658c45f226c2d524d7040d4b405c2/cmd/uninstall.go#L72
so may be it is at least a good occasion to do some refactoring just to keep things tidy?

charlie-wasp:
Resolves #170

This PR adds `getGitHooksPath()` helper as a source of truth about, where git hooks are stored

It uses `git rev-parse --git-path hooks` command under the hood (kudos to @stevelacy for the hint!), so it considers possible [overriding](https://git-scm.com/docs/githooks) default `.git/hooks` too 

This is my first PR to the repo, so please, let me know, if I can improve it in any way

### DIFF
--- a/docs/full_guide.md
+++ b/docs/full_guide.md
@@ -27,7 +27,7 @@ gem install lefthook
 ### Homebrew for macOS
 
 ```bash
-brew install evilmartians/lefthook/lefthook
+brew install lefthook
 ```
 
 ### AUR for Arch
--- a/docs/other.md
+++ b/docs/other.md
@@ -15,7 +15,7 @@ go get github.com/evilmartians/lefthook
 ### Homebrew for macOS
 
 ```bash
-brew install evilmartians/lefthook/lefthook
+brew install lefthook
 ```
 
 ### AUR for Arch


charlie-wasp:
Sorry for the mess, I realised that I've opened PR from master branch, made a new one #177

PR #177, opened 2021-05-04
`getGitHooksPath` helper function

### LINKED ISSUE #170: Custom hooks path isn't supported

charlie-wasp:
Resolves #170

This PR adds `getGitHooksPath()` helper as a source of truth about where git hooks are stored

It uses `git rev-parse --git-path hooks` command under the hood (kudos to @stevelacy for the hint!), so it considers possible [overriding](https://git-scm.com/docs/githooks) default `.git/hooks` too 

This is my first PR to the repo, so please, let me know, if I can improve it in any way

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -35,7 +35,7 @@ func TestInstallCmdExecutor(t *testing.T) {
 		"prepare-commit-msg",
 	}
 
-	files, err := afero.ReadDir(fs, getGitHooksDir())
+	files, err := afero.ReadDir(fs, getGitHooksPath())
 	assert.NoError(t, err)
 
 	actualFiles := []string{}
@@ -60,7 +60,7 @@ func TestAddCmdExecutor(t *testing.T) {
 		"pre-commit",
 	}
 
-	files, _ := afero.ReadDir(fs, getGitHooksDir())
+	files, _ := afero.ReadDir(fs, getGitHooksPath())
 	actualFiles := []string{}
 	for _, f := range files {
 		actualFiles = append(actualFiles, f.Name())
@@ -82,7 +82,7 @@ func TestAddCmdExecutor(t *testing.T) {
 		"pre-push.old",
 	}
 
-	files, _ = afero.ReadDir(fs, getGitHooksDir())
+	files, _ = afero.ReadDir(fs, getGitHooksPath())
 	actualFiles = []string{}
 	for _, f := range files {
 		actualFiles = append(actualFiles, f.Name())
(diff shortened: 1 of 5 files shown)


cmd/cmd_test.go
```diff
@@ -147,7 +147,8 @@ func presetConfig(fs afero.Fs) {
 	fs.Mkdir(filepath.Join(getRootPath(), ".lefthook/commit-msg"), defaultFilePermission)
 	fs.Mkdir(filepath.Join(getRootPath(), ".lefthook/pre-commit"), defaultFilePermission)
 
-	fs.MkdirAll(filepath.Join(getRootPath(), ".git", "hooks"), defaultFilePermission)
+  setGitHooksPath(".git/hooks")
```
DmitryTsepelev, line 150: Indent looks inconsistent but it might be a mobile-version bug

cmd/root.go
```diff
@@ -135,6 +137,25 @@ func setRootPath(path string) {
 	rootPath = strings.TrimSpace(string(outputBytes))
 }
 
+func getGitHooksPath() string {
+	return gitHooksPath
+}
+
+func setGitHooksPath(path string) {
+  gitHooksPath = filepath.Join(getRootPath(), path)
```
DmitryTsepelev, line 145: Same

charlie-wasp:
@DmitryTsepelev thanks for pointing out that formatting issues. I applied `gofmt`, so a couple of formatting changes appeared, which are not relevant to my code. Hope, it's not a problem

cmd/cmd_test.go
```diff
@@ -147,7 +147,8 @@ func presetConfig(fs afero.Fs) {
 	fs.Mkdir(filepath.Join(getRootPath(), ".lefthook/commit-msg"), defaultFilePermission)
 	fs.Mkdir(filepath.Join(getRootPath(), ".lefthook/pre-commit"), defaultFilePermission)
 
-	fs.MkdirAll(filepath.Join(getRootPath(), ".git", "hooks"), defaultFilePermission)
+	setGitHooksPath(".git/hooks")
```
Envek, line 150: Why we do need to cache hooks path (if I correctly understand what is going on)?

charlie-wasp: It's not a cache, it's more like a stub. This test is running in `cmd/` directory, and this is the way to set git hooks path to `cmd/.git/hooks`. If we use `git rev-parse --git-path hooks` here, it will return `cmd/../.git/hooks`, which is bad for tests (and breaks them actually)

Envek: Oh, ok

Envek (approved):

skryukov (approved):

Envek:
Released in [0.7.5](https://github.com/evilmartians/lefthook/releases/tag/v0.7.5). Thank you!

PR #181, opened 2021-05-07
Extract templates

skryukov:
- templates extracted
- golang version increased to support `//go:embed`
- autoinstall and run commands are using the same lookup function now


DmitryTsepelev (approved):
👍

Issue #99, opened 2019-12-24
Lefthook with Git GUI


pustovalov:
I'm currently using Husky and lint-staged with Tower, hooks with this setup work

I want to try Lefthook, but having problems

Tower
https://www.git-tower.com/mac

set hooks with this command
```
[~/kit]$ yarn lefthook install -a
yarn run v1.21.1
$ /node_modules/.bin/lefthook install -a
SYNCING lefthook.yml
SERVED HOOKS: pre-commit, prepare-commit-msg
✨  Done in 0.45s.
```
```
[~/kit/.git/hooks]$ ls -la 
total 16
drwxr-xr-x   4 pavel  staff  128 Dec 24 17:01 .
drwxr-xr-x  17 pavel  staff  544 Dec 24 17:01 ..
-rwxr-xr-x   1 pavel  staff  370 Dec 24 17:01 pre-commit
-rwxr-xr-x   1 pavel  staff  618 Dec 24 17:01 prepare-commit-msg
```

made one commit via Tower, linters have worked, but after this hooks have been deleted
```
[~/kit/.git/hooks]$ ls -la 
total 0
drwxr-xr-x   2 pavel  staff   64 Dec 24 17:03 .
drwxr-xr-x  17 pavel  staff  544 Dec 24 17:03 ..
```

unfortunately could not find a way which git commands uses Tower when committing

Source tree
https://www.sourcetreeapp.com/

Husky and lint-staged
```
git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree commit -q -F /var/folders/kj/w787fzx160130jtykf3kj1440000gn/T/SourceTreeTemp.Io8AzJ -a 
Info: can't find node in PATH, trying to find a node binary on your system
husky > pre-commit (node v13.5.0)
Stashing changes... [started]
Stashing changes... [skipped]
→ No partially staged files found...
Running tasks... [started]
Running tasks for src/**/*.{js,jsx} [started]
Running tasks for src/**/.{scss,css} [started]
Running tasks for src/**/.{scss,css} [skipped]
→ No staged files match src/**/.{scss,css}
prettier --write [started]
prettier --write [completed]
eslint --fix [started]
eslint --fix [completed]
git add [started]
git add [completed]
Running tasks for src/**/*.{js,jsx} [completed]
Running tasks... [completed]
```

Lefthook
hooks did not work when committing
```
[...]

PR #185, opened 2021-05-12
Add .lefthookrc loading

### LINKED ISSUE #178: `npm: not found` when using a git client and nvm
Since `v0.7.3` it appears Lefthook is struggling to find npm when using a git client (e.g. GitHub Desktop/SmartGit) while using `nvm`. 

_It works as expected when making a commit directly from the command line ~~or when using `v0.7.2`~~._

OS: Ubuntu

## Error
```bash
Commit failed - exit code 1 received, with output: 
Lefthook v0.7.3
⠋ waiting 
1_stylelint (SKIP. NO FILES FOR INSPECTION)
  EXECUTE > 2_eslint
 sh: 1: npm: not found
 3_prettier (SKIP BY BROKEN PIPE)
SUMMARY: (done in 0.01 seconds)
🥊  2_eslint
```

## `lefthook.yml`
```yml
pre-commit:
  piped: true
  commands:
    1_stylelint:
      tags: frontend linting styles
      glob: "*.{scss,vue}"
      run: npm exec -- stylelint {staged_files} --quiet --fix
    2_eslint:
      tags: frontend linting javascript
      glob: "*.{js,vue}"
      run: "npm exec -- eslint {staged_files} --quiet --fix"
    3_prettier:
      tags: frontend formatting styles javascript
      glob: "*.{scss,css,js,vue}"
      run: "npm exec -- prettier --check {staged_files} --write --ignore-unknown"
```

## NPM

We are currently using `nvm` to manage our `npm` installation.

```
❯ which npm
/home/lee/.nvm/versions/node/v15.14.0/bin/npm
```

If we use the standard `npm` package (`sudo apt install nodejs npm`) it works fine.

```
❯ which npm
/usr/bin/npm
```

skryukov:
Fixes #178 
Probably fixes also #99

The same workaround used in Husky https://typicode.github.io/husky/#/?id=command-not-found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+- [PR](https://github.com/evilmartians/lefthook/pull/185) Add `.lefthookrc` to fix incorrect PATH @skryukov
+- [PR](https://github.com/evilmartians/lefthook/pull/182) Support git workspaces and submodules @skryukov
+
+
 # 0.7.4 (2021-04-30)
 
 - [PR](https://github.com/evilmartians/lefthook/pull/171) Improve check for installed git @DmitryTsepelev
--- a/cmd/templates/hook.tmpl
+++ b/cmd/templates/hook.tmpl
@@ -8,6 +8,10 @@ if [ -t 1 ] ; then
   exec < /dev/tty ; # <- enables interactive shell
 fi
 
+if [ -f ~/.lefthookrc ]; then
+  source ~/.lefthookrc
+fi
+
 dir="$(cd "$(dirname "$(dirname "$(dirname "$0")")")" >/dev/null 2>&1 || exit ; pwd -P)"
 
 call_lefthook()
--- a/spec/fixtures/pre-commit
+++ b/spec/fixtures/pre-commit
@@ -8,6 +8,10 @@ if [ -t 1 ] ; then
   exec < /dev/tty ; # <- enables interactive shell
 fi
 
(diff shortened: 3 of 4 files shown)


Envek:
Why we need one more config file? Why we can't just add some key to [`lefthook-local.yml`](https://github.com/evilmartians/lefthook#local-config) for example?

SAGARSURI:
Any update on this?

skryukov:
So the problem is - we need to load `PATH` using shell script only (in the case of npm-based Lefthook installation). Parsing a full-featured yaml file is overkill. We can `grep` file to find line by key name, but what if it will be multiline value? In that case, `.lefthookrc` seemed like a good enough solution.

@Envek wdyt?

Envek:
In #178 lefthook itself is being found correctly, parses Lefthook YAML config and fails to execute commands defined there due to the incomplete PATH set by GUI clients. So there is no need to execute anything in the hook file before invoking Lefthook itself.

skryukov:
@Envek in #178 Lefhook installed as a local dependency via npm, so I think it is found by this line https://github.com/evilmartians/lefthook/blob/master/cmd/templates/hook.tmpl#L20

But `npx lefthook`, `yarn lefthook`, and even `bundle exec lefthook` won't work as far as I understand.

lm93547:
Any update on this

JosephNK:
this update need

Krunal-K-SimformSolutions:
Any update? how much time does it take to merge this PR because I have the same issue?

PR #187, opened 2021-05-15
Add ability to skip summary output. Also support a LEFT_QUIET env variable

washtubs:
Decided to tackle some further issues raised [here](https://github.com/evilmartians/lefthook/issues/78#issuecomment-658333189). There was mention of using a global environment variable so you didn't need to rely on the yaml having it configured. This is personal preference after all.

*My* personal preference is I like it very quiet. So I made it so you could get rid of the summary line as well.

### DIFF
--- a/docs/full_guide.md
+++ b/docs/full_guide.md
@@ -281,7 +281,7 @@ If you need to extend config from some another place, just add top level:
 ```yml
 # lefthook.yml
 
-extends: 
+extends:
   - $HOME/work/lefthook-extend.yml
   - $HOME/work/lefthook-extend-2.yml
 ```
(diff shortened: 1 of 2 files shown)


docs/full_guide.md
```diff
@@ -469,6 +469,28 @@ source_dir: ".lefthook"
 source_dir_local: ".lefthook-local"
 ```
 
+## Mange verbosity
```
Envek, line 472: Typo:
```suggestion
## Manage verbosity
```

Envek (commented):
Looking good to me, except one typo.

Also it would be awesome to have 2 commits: one with option for summary output suppresion and one with environment variable addition.

washtubs:
Done and done!

Envek (approved):

Envek:
Awesome, thank you!

Issue #136, opened 2020-07-14
Install Lefthook using a package manager in Windows


SAGARSURI:
Is there a way to install lefthook on Windows using `chocolatey` package manager or `winget`? I want to write a bash script to automate this process of installing lefthook on my machine and other teammates.

Arkweid:
@SAGARSURI Hi, you could add lefthook in any package manager you want, like here for python: https://github.com/life4/lefthook

SAGARSURI:
Hey @Arkweid , Is it possible to add the `.exe` file in the asset of the release notes? I am planning to open up a PR for the `winget` package manager. They expect to mention the type of installer while submitting the PR.

Arkweid:
Oke, I will add unzipped file in next release.

SAGARSURI:
Thank you so much! @Arkweid

SAGARSURI:
I have added this to my JIRA ticket and will resume work once this is done. 😄

ahuglajbclajep:
On Windows 1803 or later, you can use Unix `curl` with `curl` (in CMD) or `curl.exe` (in PowerShell).
So when unzipped files are added to the release, you can install `lefthook` with the same commands in both Windows and Unix (although you have to change the download URL and set the `PATH`) 🎉

bbodenmiller:
It does not appear `exe` ever started getting published. @Envek could this be added for next release?

SAGARSURI:
Thank you so much @Envek

aminya:
This is counterintuitive. You should probably use a better package manager. For example, Chocolatey is able to unzip files without any issues.

SAGARSURI:
Why should we use Chocolatey when Microsoft provide a default package manager i.e `winget`. It provides better security compared to other 3rd party package manager.

aminya:
> Why should we use Chocolatey when Microsoft provide a default package manager i.e `winget`. It provides better security compared to other 3rd party package manager.

I'm very sorry to see this attitude toward community-driven open-source projects. A 3rd party project is not insecure by default, and Microsoft projects are **not** any more secure. Instead, I suggest you decide based on the actual technical differences. For example, if a package manager lacks its basics such as installing zip files or if it allows decentralized packages.

_P.S: since this is offtopic, I will not continue this discussion after this comment._

Envek:
We want to support as many installations methods as possible (however, using as least effort as possible). So, let's add support to both (pull requests are welcome)

[...]

PR #188, opened 2021-05-15
Npm: download the required binaries during installation

aminya:
- The binaries are now downloaded upon installation. This reduces the package size significantly, which is very important for open-source repositories.

- Fixes finding the binaries on Windows inside the hooks. Previously the code fell back to using `npx`

Closes #43


.npm/postinstall.js
```diff
@@ -1,10 +1,69 @@
...
+  const version = require("./package.json").version
+
+  return `https://github.com/evilmartians/lefthook/releases/download/v${version}/lefthook_${version}_${downloadOS}_${arch}${extension}`
+}
+
+const { DownloaderHelper } = require("node-downloader-helper")
+const path = require("path")
+
+async function downloadBinary() {
+  // TODO zip the binaries to reduce the download size
```
aminya, line 52: I noticed that you have uploaded unzipped binaries directly to the GitHub releases. We should probably upload zipped binaries to reduce the download time.

aminya: We can also make a separate set of zip files to prevent conflict with what people want in #136. This would be probably a better way as we can fulfill the needs of all parties.

aminya: Make a PR in #189

.npm/postinstall.js
```diff
@@ -1,10 +1,69 @@
...
+  const extension = path.extname(downloadURL).slice(1)
+  const fileName = `lefthook.${extension}`
+  const binDir = path.join(__dirname, "bin")
+  const dl = new DownloaderHelper(downloadURL, binDir, {
+    fileName,
+    retry: { maxRetries: 5, delay: 50 },
+  })
+  dl.on("end", () => console.log("lefthook binary was downloaded"))
+  await dl.start()
+  return path.join(binDir, fileName)
```
Envek, line 63: On Linux and MacOS downloaded file doesn't have access right to be executed. To fix this, you need to set executable bit here. You can use code like this:

```js
const { chmod } = require("fs/promises");
await chmod(pathToDownloadedBinary, "755");
```

Where `755` is the same access rights as for `bin/index.js` (`755` is the same as `rwxrwxr-x` — owner and group can read, write, and execute, but others can only read and execute)

Envek (changes requested):
Hello! Thank you for your contribution and sorry for the late review!

Recently we had discussions on whether we should change packaging to download binaries on install or not and finally decided that we should. So your pull request is just in time!

However, there are some issues that should be addressed first:

 1. Download is broken on Linux and MacOS: file has wrong name due to incorrect filename extension detection and missing executable bit.
 2. In case if download failed, error message is unclear:

    ```
    [5/5] Rebuilding all packages...
    error /home/envek/test/node_modules/@arkweid/lefthook: Command failed.
    Exit code: 1
    Command: node postinstall.js
    Arguments: 
    Directory: /home/envek/test/node_modules/@arkweid/lefthook
    Output:
    events.js:292
          throw er; // Unhandled 'error' event
          ^
    
    Error: Response status was 404
        at ClientRequest.<anonymous> (/home/envek/evl.ms/test/node_modules/node-downloader-helper/dist/index.js:1:7153)
        at Object.onceWrapper (events.js:422:26)
        …
    ```

     To help users to understand what is going wrong (and to help us debug it), error message (`Response status was 404` in this example) should be prefixed with details about what and when failed. Something like this:

    ```
    Failed to download lefthook_0.7.5_Linux_x86_64: Response status was 404 while fetching https://github.com/evilmartians/lefthook/releases/download/v0.7.5/lefthook_0.7.5_Linux_x86_64
    ```

.npm/bin/index.js
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env node
 
 var spawn = require('child_process').spawn;
-const { getExePath } = require('../get-exe');
+const path = require("path")
+const exePath = path.join(__dirname, `lefthook${["win32", "cygwin"].includes(process.platform) ? ".exe" : ""}`)
```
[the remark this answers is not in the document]
aminya: Fixed in 
https://github.com/evilmartians/lefthook/pull/188/commits/12645c2344d31294169454245b5eff3536cfa463

.npm/postinstall.js
```diff
@@ -1,10 +1,69 @@
...
+}
+
+const { DownloaderHelper } = require("node-downloader-helper")
+const path = require("path")
+
+async function downloadBinary() {
+  // TODO zip the binaries to reduce the download size
+  const downloadURL = getDownloadURL()
+  const extension = path.extname(downloadURL).slice(1)
+  const fileName = `lefthook.${extension}`
```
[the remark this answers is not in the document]
aminya: Fixed in https://github.com/evilmartians/lefthook/pull/188/commits/601b7b55bfb006c53e2c58e4b4ed70902bb280be

.npm/postinstall.js
aminya: Fixed in https://github.com/evilmartians/lefthook/pull/188/commits/4c46696928589ee3e06edd40031ed012a30f9206

aminya:
Thank you. I addressed the comments.

.npm/postinstall.js
```diff
@@ -1,10 +1,78 @@
...
+  const binDir = path.join(__dirname, "bin")
+  const dl = new DownloaderHelper(downloadURL, binDir, {
+    fileName,
+    retry: { maxRetries: 5, delay: 50 },
+  })
+  dl.on("end", () => console.log("lefthook binary was downloaded"))
+  try {
+    await dl.start()
+  } catch(e) {
+    console.error(`Failed to download ${fileName} while fetching ${downloadURL}`)
```
Envek, line 69: Same: in yarn, no `console.error` output isn't displayed to user terminal. Do you see it?

Envek (changes requested):

Envek (approved):

.npm/postinstall.js
```diff
@@ -1,10 +1,76 @@
...
+  const extension = ["win32", "cygwin"].includes(process.platform) ? ".exe" : ""
+  const fileName = `lefthook${extension}`
+  const binDir = path.join(__dirname, "bin")
+  const dl = new DownloaderHelper(downloadURL, binDir, {
+    fileName,
+    retry: { maxRetries: 5, delay: 50 },
+  })
+  dl.on("error", (e) => {
+    e.message = `Failed to download ${fileName}: ${e.message} while fetching ${downloadURL}`
+    throw e
```
aminya, line 67: Does throwing the error here prevents retrying? What I had here allowed the code to retry downloading.

Envek: Oh, this is good question.

bbodenmiller:
FYI this may break some workflows. Some organizations run a proxy for npm packages but block other internet access or disallow direct downloads by policy.

Envek:
Yes, we [know that](https://github.com/evilmartians/lefthook/pull/43#issuecomment-591440728). In that case other installation methods can be used instead.

I wish that NPM supported platform-specific packages, but in current form it is already way too heavy,

.npm/postinstall.js
Envek: Looks like it actually prevent retrying. But I believe [it is a bug](https://github.com/hgouveia/node-downloader-helper/issues/57).

cmd/templates/hook.tmpl
```diff
@@ -14,24 +14,24 @@ call_lefthook()
 {
   if lefthook -h >/dev/null 2>&1
   then
-    eval $1
-  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/lefthook"
+    eval lefthook $1
```
Envek, line 17: Should we also add `{{.Extension}}` here (and in `if` above)? I think yes, because if someone will install Lefthook system-wide in Windows, it can be called only via `lefthook.exe` then.
```suggestion
    eval lefthook{{.Extension}} $1
```

.npm/postinstall.js
aminya: I think we should revert this change. What I had here allowed retrying, and also printed the actual error message.

Envek:
I extracted everything about executable extension into separate commit and pushed it to master, so here we can focus on NPM package (no action needed).

.npm/postinstall.js
Envek: Well, the problem is that printing doesn't work (at least with yarn). I will experiment with error throwing again later.

aminya: You could replace this `throw e` with `console.error(e)`, and then revert what I had to throw the last error.
```suggestion
    console.error(e)
```

bbodenmiller:
> Yes, we [know that](https://github.com/evilmartians/lefthook/pull/43#issuecomment-591440728). In that case other installation methods can be used instead.
> 
> I wish that NPM supported platform-specific packages, but in current form it is already way too heavy,

Perhaps a net install (this MR) vs full install (current package) could be maintained then. Alternatively it appears there is options to handle this using OS dependencies: https://stackoverflow.com/questions/15176082/npm-package-json-os-specific-dependency

aminya:
You could make a fork, and add proxy support similar to [prebuild-install](https://github.com/prebuild/prebuild-install/blob/master/proxy.js)

bbodenmiller:
> You could make a fork, and add proxy support similar to [prebuild-install](https://github.com/prebuild/prebuild-install/blob/master/proxy.js)

Proxy support does not address organizations that forbid external downloads by policy and/or block external internet access other than via npm mirror.

Do what you want, I'm just telling you have this will break some workflows and it appears I'm not the first to raise this concern.

aminya:
If the company blocks the internet, it should mock it using its proxy server. `https://github.com/evilmartians/lefthook/releases/download` URL can be simply redirected to a local server that hots the binaries. This is probably similar to how that company is mocking the URLs for the packages that use [prebuild-install](https://www.npmjs.com/package/prebuild-install) and [node-pre-gyp](https://www.npmjs.com/package/@mapbox/node-pre-gyp).

.npm/postinstall.js
Envek: `console.error` output isn't visible when yarn is used to install package.

Envek:
We probably would like to support two packages (one with bundled executables and one that downloads them), but it requires changing how hook scripts invoke lefthook binary.

.npm/postinstall.js
aminya: Then there is no way to retry with Yarn. I recommend throwing the last error for those on Yarn and let the algorithm work normally for others.

Envek: Reverted my commit

I couldn't make yarn to print error message in a descriptive way I want because yarn is exiting on the first `error` event thrown (and I insist that this is caused by the bug in node-download helper, see https://github.com/hgouveia/node-downloader-helper/issues/57#issuecomment-848817013).

I give up. Let it be so.

Envek:
Thank you! We will release it soon. Maybe next week.

aminya:
Thanks! 🎉

privatenumber:
Since this fetches binaries from GitHub, I want to raise the concern that this could interfere with environments that rely on a npm proxy registry (as originally raised [here](https://github.com/evilmartians/lefthook/pull/43#issuecomment-591440728)).

npm proxies are not only used in systems that are firewalled/have limited internet, but also in many (enterprise) environments that proxy npm in case npm going down doesn't block CIs (eg. deploys).

I think this can be easily addressed by publishing the binaries to npm and downloading from npm instead of GitHub.

aminya:
> Thank you! We will release it soon. Maybe next week.

Any plan to release this?

aminya:
@Envek

Envek:
Sorry, hadn't a chance to prepare a release (it is pretty big compared to previous ones), so it is postponed until I will have a day or two to polish things out. Can't say any specific date or time now.

Envek:
It finally was released in [0.8.0](https://github.com/evilmartians/lefthook/releases/tag/v0.8.0) and now available as [`@evilmartians/lefthook-installer`](https://www.npmjs.com/package/@evilmartians/lefthook-installer) package.

Thank you very much for you contribution and sorry for such a long delay.

PR #189, opened 2021-05-15
Include archived binaries in the releases

aminya:
This brings back the archived binaries. This makes downloading the binaries faster in #188. This also means that we can fulfill the needs of all parties (those who requested raw binaries in the release and us who are writing an automated code to download the binaries for the npm package).

To test this:
```
goreleaser --snapshot --skip-publish --rm-dist
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,16 @@ archives:
       linux: Linux
       386: i386
       amd64: x86_64
+  - id: lefthook-gz
+    format: gz
+    files:
+    - none*
+    replacements:
+      windows: Windows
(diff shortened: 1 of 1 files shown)


Envek:
I have contradicting feelings about this. From the one hand I think that we don't need excess files in GitHub releases, but from the other hand GitHub doesn't compress downloaded files even if asked with `Accept-Encoding` header (I tested with ` curl --compressed --verbose --location`) so it may worth it.

aminya:
In my opinion, not only It doesn't hurt to provide more files in the release, but also it helps many of us. So, I am all for including these.

PR #184, opened 2021-05-09
Rewrite npm's scripts in JavaScript

aminya:
- This removes the dependency on `sh` for a Node env, which is not available on Windows. 
- It also speeds up the startup time because there is no `sh` script anymore to slow things down. Nodejs itself spawns the executable directly.

Fixes #183 
Related to #16

PR #190, opened 2021-05-15
Fix finding lefthook binary under node_modules

aminya:
In #184, the lefthook sh file was removed, but the Node package managers create one automatically under  `node_modules/.bin`. This fixes finding binaries under node_modules no matter what package manager is used.

It also fixes finding binaries for the cases where `$dir` has spaces in it (allowed on Windows)


aminya:
This doesn't seem to work. I made a commit which supersedes this PR:
https://github.com/evilmartians/lefthook/pull/188/commits/1b06157438378ac3118f8e64ba1b84831d1c8008

Issue #122, opened 2020-03-24
Documentation to work with commitlint?


radiosilence:
Unfortunately the commitlint documentation only has these instructions:

```json
{
  "husky": {
    "hooks": {
      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
    }
  }
}
```

And it's unclear whether lefthook exposes whatever `HUSKY_GIT_PARAMS` are in such a way that it works with commitlint.

Arkweid:
Hi @radiosilence
It is impossible to store documentation for every tool in the world. Check the full guide and if you have a questions feel free to ask.

Maybe that what a you looking for?
https://github.com/Arkweid/lefthook/blob/master/docs/full_guide.md#bash-script-example

muuvmuuv:
I use it like this:

```yml

    lint-commit:
      run: npx commitlint --from HEAD~1 --to HEAD
```

tillsanders:
@muuvmuuv Somehow, this doesn't work for me. I also read, that the `pre-commit` hook should not work for this usecase. Apparently, the `commit-msg` hook should be used instead. Though this won't work for me as well :/

tillsanders:
@Arkweid I took a look at your example using the bash script. I'm trying to figure out, how this could be used with a tool like commitlint without using a separate script file. Is there a way to access the commit message in the run command? Maybe using an environment variable or an argument?

Arkweid:
For simple `run` lefthook not pass arguments from git.
If you need just a commit message, so as a mentioned above you could use a various git commands like [git show](https://git-scm.com/docs/git-show)

muuvmuuv:
Oh, sorry, noticed that my above does not work, but this one does:

```yaml
commit-msg:
  commands:
    lint-commit-msg:
      run: npx commitlint --edit
```

```
  --edit, -e             read last commit message from the specified file or fallbacks to ./.git/COMMIT_EDITMSG
```

paambaati:
I have a similar use-case with `git-notify`, where I need to pass Git parameters to `post-merge`, `post-rewrite` and `post-checkout` hooks, with the params being dynamic — https://github.com/jevakallio/git-notify#installing-hooks-by-any-other-means

Has someone had success doing this?

muuvmuuv:
@paambaati where is this related to Commitlint? Post a new issue for this please.

PawFV:
> ```yaml
> commit-msg:
>   commands:
>     lint-commit-msg:
[...]

Issue #110, opened 2020-02-03
Commitlint


binjospookie:
Hi!
I want to use Lefthook with Commitlint on pre-commit.
In Commitlint's docs I found `"commit-msg": "commitlint -E HUSKY_GIT_PARAMS"` (https://github.com/conventional-changelog/commitlint#getting-started). But Lefthook hasn't `HUSKY_GIT_PARAMS` of course :)

What should I do to solve this problem?

Arkweid:
Try it like usual executable you have:
commit-msg:
  commands:
    commitlint:
      run: echo "wololo" | commitlint

Or try it just like script:
https://github.com/Arkweid/lefthook/blob/master/docs/full_guide.md#bash-script-example

binjospookie:
Thank you! I'll try and update this issue with new information.

binjospookie:
I modify your script with type-enum from https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum

Looks like:
```bash
INPUT_FILE=$1
START_LINE=`head -n1 $INPUT_FILE`
PATTERN="^(build|ci|docs|feat|fix|perf|refactor|revert|style|test)+: "
if ! [[ "$START_LINE" =~ $PATTERN ]]; then
  echo -e "Invalid commit message.\nCommit message must start with '(build|ci|docs|feat|fix|perf|refactor|revert|style|test):'"
  exit 1
fi
```

Thank you for help!

PR #201, opened 2021-05-26
Add commitlint example in full_guide

mihaic195:
- Relates to this issue: https://github.com/evilmartians/lefthook/issues/122 The solution provided here doesn't stop the process of committing.
- Also mentioned here https://github.com/evilmartians/lefthook/issues/110 but I find that is not the right solution as it doesn't use commitlint.

### DIFF
--- a/docs/full_guide.md
+++ b/docs/full_guide.md
@@ -224,7 +224,7 @@ When you try to commit `git commit -m "haha bad commit text"` script `template_c
 
 ## Bash script example with Commitlint
 
-Let's create a bash script to check commit templates `.lefthook/commit-msg/commitlint.sh`:
+Let's create a bash script to check conventional commit status `.lefthook/commit-msg/commitlint.sh`:
 
 ```bash
 INPUT_FILE=$1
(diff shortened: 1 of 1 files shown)


docs/full_guide.md
```diff
@@ -222,6 +222,34 @@ commit-msg:
...
+Let's create a bash script to check conventional commit status `.lefthook/commit-msg/commitlint.sh`:
+
+```bash
+INPUT_FILE=$1
+START_LINE=`head -n1 $INPUT_FILE`
+ERROR=`echo $START_LINE | npx commitlint --color`
+if ! [[ $? == 0 ]]; then
+  echo "${ERROR}"
+  exit 1
+fi
```
Envek, line 236: What is the purpose of this `if` block? I believe that if script consists of this single line:
```sh
echo $(head -n1 $1) | npx commitlint --color
```
it will behave the same (printing error and exiting with non-zero)

aminya, line 236: A separate shell script is not needed anymore after
https://github.com/evilmartians/lefthook/commit/8defa7e2a3bd4385d3a296056c232c28183328dc

mihaic195: Indeed, you are right. My bash skills are not great 😅
Updated.

mihaic195: Can you please elaborate? I don't quite follow...

mrexox:
Great job! This is pretty useful!

PR #205, opened 2021-06-03
Docs micro-update

HellSquirrel:
Added small updates to docs

Moved the npm global installation example down.
Added tsx and jsx extensions


Envek:
Thanks!

Issue #206, opened 2021-06-04
Subcommands return `status 128`


pwinckles:
I downloaded the Linux x86_64 binary release of `0.7.6` and all of the subcommands return `exit status 128`. I then tried downloading `0.7.5` and `0.7.4`. `0.7.5` behaves the same way, but `0.7.4` works as expected.

Envek:
That's right, we have a regression (or reincarnation of #90) here. Thanks for reporting.

PR #209, opened 2021-06-09
do not init git config on help and version cmds

pwinckles:
This PR resolves https://github.com/evilmartians/lefthook/issues/206 and https://github.com/evilmartians/lefthook/issues/90.

The root cause is that `git` commands were being run for all subcommands, instead of just the ones that needed them. This PR changes when the git config is resolved so that the problematic commands are not run for `version` and `help`.

Additionally, it slightly improves the error message when you invoke a subcommand that does need to be run from within a git repo. Previously, it would produce a cryptic `exit status 128` response.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -25,6 +25,8 @@ const (
 	configExtensionPattern  string      = ".*"
 	defaultFilePermission   os.FileMode = 0755
 	defaultDirPermission    os.FileMode = 0666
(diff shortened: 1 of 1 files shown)


skryukov (approved):

Envek (approved):

Envek:
Released in [0.7.7](https://github.com/evilmartians/lefthook/releases/tag/v0.7.7)

PR #213, opened 2021-06-24
Abort hook if lefthook was not found

roman-ih:


skryukov:
What if we add some flag to the `install` and `add` commands to provide an opt-in option for this behavior?

@Envek wdyt?

roman-ih:
I don't really have bandwidth to spend much time on this but yeah, I agree, ideally I'd love if there was a more precise way to configure lefthook.

As an idea, having an option in `lefthook.yml` to control this would be great. Also, why not to add an option to install lefthook automatically when one does `bundle install` for the first time in freshly cloned repo? - similar to how husky works in nodejs.

mrexox:
Hey! Sorry for the long delay. It feels like changing current behavior is a big change for lefthook. So, I'd like to suggest an option.

What do you think of having a **global config option** for that, like `ensure: true` or `mandatory: true`? Git hooks will have `exit 1` if this option is set to `true`. So, current lefthook users won't have problems when upgrading, but you can make your configuration stricter and require lefthook to be available (through PATH or other supported options).

mrexox:
This issue was resolved with `assert_lefthook_installed` option that is set to `false` by default. You can set it to `true` if you want to make sure lefthook is not just skipped if it was ever installed in your git hooks.

https://github.com/evilmartians/lefthook/pull/533

PR #218, opened 2021-08-03
node: fix docs to actually refer to the correct package

ThisIsMissEm:
It turns out that the lefthook package is actually a fake, and shouldn't be run, instead you'd want this change.

### DIFF
--- a/docs/node.md
+++ b/docs/node.md
@@ -33,7 +33,7 @@ pre-commit:
 
 ## Test it
 ```bash
-npx lefthook install && npx lefthook run pre-commit
+npx @arkweid/lefthook install && npx @arkweid/lefthook run pre-commit
 ```
 
 ### More info


ThisIsMissEm:
The wiki also needs to be updated to mention how to uninstall husky: https://github.com/evilmartians/lefthook/wiki/Migration-from-husky

Specifically the `git config --unset core.hooksPath` because husky messes with that.

Envek:
Awesome! Thank you!

Interesting that `npx lefthook run pre-commit` works on my machine because I have lefthook installed from system package manager:
```sh
$ npx which lefthook                  
/usr/bin/lefthook
```

Envek:
Also edited wiki. Thanks again!

ThisIsMissEm:
Yeah, I'm not sure exactly what npx is doing there; it was working on my machine, but I thought I was getting trolled by the owner of the lefthook package on npm because weird husky stuff was happening (turned out they'd messed with the gitHooks path)

ThisIsMissEm:
I've also submitted a ticket to npm about potentially getting the lefthook package name unscoped, as it's currently unused and there's lots that references it incorrectly

Issue #219, opened 2021-08-05
add option for disabling the spinner


pwinckles:
It would be nice to have an option to disable the spinner similar to how colors can be disabled. The spinner produces output like the following in some environments:

```
[K⠋ waiting
[K⠙ waiting
[K⠹ waiting
[K⠸ waiting
[K⠼ waiting
[K⠴ waiting
[K⠦ waiting
[K⠧ waiting ....
```

mrexox:
Hey! I've just added this option - [`no_tty`](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#no_tty) and I am going to release it soon.

Issue #208, opened 2021-06-08
The command output is suppressed while in progress


styx:
Version 0.7.3 has changed how the output is produced to the console. In the result i.e. when running `rspec` with progress bar you won't see it and only spinner from lefthook. And later 100% of progress bar. Also this is not comfortable when some tests are broken. You will know only in the end.

I believe this change introduced this regression https://github.com/evilmartians/lefthook/pull/130/files#diff-8146f8148ccbf6711d65f532f6ab9a7c8dfbdc3960c7ffc974f17d6d224dd349L260

ShPakvel:
Agree. I suggest to return back ability to see logs from all invoked commands. At least it can be done via configuration or lets say env var. Though I feel like logs should be default behavior, and spinner set via configuration (or e.g. env var).

Envek:
We have many controversal views on this behavior, so yes, it should be configurable.

For now, I can only recommend using `--fail-fast` option for rspec so it fails on the first broken test.

iSuslov:
What if I have lefthook running in my CI. It outputs almost infinite `waiting` line during run. Look at the screenshot.
<img width="1021" alt="Screen Shot 2022-05-19 at 7 52 10 PM" src="https://user-images.githubusercontent.com/7628774/169422507-d17ce575-45b3-4ef5-869f-94fae61b4038.png">

**Is it really a desired behavior?**   

Anyway I want to see lefthook output in realtime. 
Why should I wait for my target to complete before I can see some debugging info? @Envek @Arkweid

JoyceBabu:
My git client shows only the last 100 lines of output, and I can see only the spinner output.

styx:
@mrexox it's unfortunate to say that, but behaviour is still different from what was in 0.7.2 version

0.7.2 https://youtu.be/rHzHpPlgIQc
1.2.5 with --no-tty https://youtu.be/pRHVMrwj_MA

mrexox:
Hmm, I see. @styx, what do you think about `-f/--follow` option which would bring comment output to stdout like it did in 0.7.2? It is going to be a mess when used with [`parallel`](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#parallel) option but I guess it would fit your needs.

I think we can switch to follow-mode implicitly for hooks without `parallel` option, but I need to make sure it doesn't break anything else.

styx:
@mrexox sounds good to me

PR #220, opened 2021-08-05
disable the spinner for non-tty output; add option to explicitly disa…

pwinckles:
A possible resolution to https://github.com/evilmartians/lefthook/issues/219

1. The spinner is now disabled by default for non-tty output
2. The spinner can be disabled with the `--no-spinner` flag
3. The spinner can be configured with `spinner` in `lefthook.yml`

Not the prettiest impl, but it should work.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -32,6 +32,7 @@ Change working directory or initialize new repository with 'git init'.`
 var (
 	Verbose              bool
 	NoColors             bool
+	NoSpinner            bool
 	rootPath             string
 	gitHooksPath         string
 	originConfig         *viper.Viper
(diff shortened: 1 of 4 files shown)


Envek:
First of all, thank you for the pull request! Animated spinner apparently isn't fits everyone's needs in every circumstance, so we absolutely need a way to disable it.

However, from `--no-spinner` option (or `spinner: false` config key) I would expect that it not only removes animation, but also brings back output of commands being executed (see https://github.com/evilmartians/lefthook/issues/208 for additional context).

Let's discuss, which behavior is least surprising and most expected by users

pwinckles:
I'm not sure if/when I'll have a chance to attempt that, so I'm just going to close this PR

palkan:
> I would expect that it not only removes animation...

Sounds reasonable, there could appear other interactive features; and I would call it `--non-tty` or smth like that.

> ... , but also brings back output of commands being executed

Not sure about this part. Displaying commands and outputs looks more like a verbose mode to me.

PR #224, opened 2021-09-02
Allows lefthook to work when node_modules is not in root folder for npx

spearmootz:
npx will automatically install the module if not present. yarn does not do this and complains when the package.json is not found.

### DIFF
--- a/cmd/templates/hook.tmpl
+++ b/cmd/templates/hook.tmpl
@@ -21,7 +21,7 @@ call_lefthook()
   elif bundle exec lefthook -h >/dev/null 2>&1
   then
     bundle exec lefthook $1
-  elif npx lefthook -h >/dev/null 2>&1
+  elif npx @arkweid/lefthook -h >/dev/null 2>&1
   then
     npx lefthook $1
   elif yarn lefthook -h >/dev/null 2>&1


spearmootz:
the reason for this is that my project has a main folder and a /src folder.  the package.json and node_modules is found in the src folder and therefore npx lefthook is not found but it is also not automatically installed because it is missing the @arkweid

Envek:
Thank you!

spearmootz:
can this be released?

cmd/templates/hook.tmpl
```diff
@@ -21,7 +21,7 @@ call_lefthook()
   elif bundle exec lefthook -h >/dev/null 2>&1
   then
     bundle exec lefthook $1
-  elif npx lefthook -h >/dev/null 2>&1
+  elif npx @arkweid/lefthook -h >/dev/null 2>&1
   then
     npx lefthook $1
```
PikachuEXE, line 26: Should this line be updated too?

Envek: It may be updated and may be not updated. `npx` seems to search for binaries both in `node_modules` and in system `PATH`, so if `if` statement above if passes then there will be `lefthook` binary in `node_modules`.

spearmootz: no, that wont work. maybe both should be checked.  this is for the case when the module is not in the root of the git project. so doing npx lefthook wont run it.

spearmootz: Oh yeah, my bad. The line below also needs to be npx @arkweid/lefthook

Envek:
Released in [0.7.7](https://github.com/evilmartians/lefthook/releases/tag/v0.7.7)

PR #234, opened 2021-09-28
Cleanup hook template

### LINKED ISSUE #232: Issue with hook template binary lookup / released binary on MacOS
In https://github.com/evilmartians/lefthook/blob/master/cmd/templates/hook.tmpl#L18-L20

It tried to look for `./node_modules/@arkweid/lefthook`

But there are only multiple executable files inside folders for different platforms
image
image

I have no idea how to fix this

I am using SourceTree but it's able to detect base folder via 
https://github.com/evilmartians/lefthook/blob/master/cmd/templates/hook.tmpl#L11

But I am still unable to use due to this issue

PikachuEXE:
This was for #232 

But as @aminya point out `npx` already replaced `node_modules/@arkweid/lefthook/bin/lefthook`

There is code left which is useless at the point and can be cleanup

So this PR is updated to clean that code up

### DIFF
--- a/cmd/templates/hook.tmpl
+++ b/cmd/templates/hook.tmpl
@@ -15,9 +15,6 @@ call_lefthook()
   if lefthook{{.Extension}} -h >/dev/null 2>&1
   then
     eval lefthook{{.Extension}} $@
-  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}}"
-  then
-    eval "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}} $@"
   elif bundle exec lefthook -h >/dev/null 2>&1
   then
     bundle exec lefthook $@


cmd/templates/hook.tmpl
```diff
@@ -15,9 +15,9 @@ call_lefthook()
   if lefthook{{.Extension}} -h >/dev/null 2>&1
   then
     eval lefthook{{.Extension}} $@
-  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}}"
+  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/index.js"
   then
-    eval "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}} $@"
+    eval "$dir/node_modules/@arkweid/lefthook/bin/index.js $@"
```
aminya, line 20: This doesn't work. The test is always true

aminya (changes requested):

cmd/templates/hook.tmpl
PikachuEXE: You mean `test -f "$dir/node_modules/@arkweid/lefthook/bin/index.js"` always return `true`?

aminya: Yes. There is no point in the test. Also, you can't eval JavaScript directly. In the new version, there is no need for `index.js`.

PikachuEXE: `test -f "$dir/node_modules/@arkweid/lefthook/bin/index.js"` only returns true if `lefthook` is installed via npm/yarn
When uninstalled it would only echo `Can't find lefthook in PATH`
If `lefthook` installed via Homebrew, the `test -f` line won't even be called

<details>
<summary>My custom `pre-commit` file</summary>
```sh
#!/bin/sh

if [ "$LEFTHOOK" = "0" ]; then
  exit 0
fi

if [ -t 1 ] ; then
  exec < /dev/tty ; # <- enables interactive shell
fi

dir="$(git rev-parse --show-toplevel)"

call_lefthook()
{
  if lefthook -h >/dev/null 2>&1
  then
    eval lefthook $@
  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/index.js"
  then
    eval "$dir/node_modules/@arkweid/lefthook/bin/index.js $@"
  elif bundle exec lefthook -h >/dev/null 2>&1
  then
    bundle exec lefthook $@
  elif npx lefthook -h >/dev/null 2>&1
  then
    npx lefthook $@
  elif yarn lefthook -h >/dev/null 2>&1
  then
    yarn lefthook $@
  else
    echo "Can't find lefthook in PATH"
  fi
}



call_lefthook "run pre-commit $@"

```
</details>


<details>
<summary>My custom `pre-commit` file with `echo` on each "branch"</summary>
```sh
#!/bin/sh

if [ "$LEFTHOOK" = "0" ]; then
  exit 0
fi

if [ -t 1 ] ; then
  exec < /dev/tty ; # <- enables interactive shell
fi

dir="$(git rev-parse --show-toplevel)"

call_lefthook()
{
  if lefthook -h >/dev/null 2>&1
  then
    echo "pure lefthook route"
    eval lefthook $@
  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/index.js"
  then
    echo "node_modules/@arkweid/lefthook/bin/index.js route"
    eval "$dir/node_modules/@arkweid/lefthook/bin/index.js $@"
  elif bundle exec lefthook -h >/dev/null 2>&1
  then
    echo "bundle exec route"
    bundle exec lefthook $@
  elif npx lefthook -h >/dev/null 2>&1
  then
    echo "npx route"
    npx lefthook $@
  elif yarn lefthook -h >/dev/null 2>&1
  then
    echo "yarn route"
    yarn lefthook $@
  else
    echo "Can't find lefthook in PATH"
  fi
}



call_lefthook "run pre-commit $@"

```
</details>

PikachuEXE: You are saying `npx lefthook` is already using `node_modules/@arkweid/lefthook/bin/index.js`.
So this check is redundant and should be removed?

aminya: Yes. This PR doesn't solve anything. 

index.js is used when running npx lefthook
https://github.com/evilmartians/lefthook/blob/6c84d855dbb7874d9a82713c15441af940851a44/.npm/package.json#L7

PikachuEXE: I will update this PR to just remove this useless branch then

Something I just noticed:
In https://github.com/evilmartians/lefthook/blob/6c84d855dbb7874d9a82713c15441af940851a44/cmd/templates/hook.tmpl#L24-L29

I see `@arkweid/lefthook` and `lefthook` used at the same time
Is it supposed to use `@arkweid/lefthook` only?

[`lefthook`](https://www.npmjs.com/package/lefthook) != [`@arkweid/lefthook`](https://www.npmjs.com/package/@arkweid/lefthook) on npmjs.com

cmd/templates/hook.tmpl
```diff
@@ -15,9 +15,6 @@ call_lefthook()
   if lefthook{{.Extension}} -h >/dev/null 2>&1
   then
     eval lefthook{{.Extension}} $@
-  elif test -f "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}}"
-  then
-    eval "$dir/node_modules/@arkweid/lefthook/bin/lefthook{{.Extension}} $@"
```
aminya, line 20: The executable can be spawned directly. It is slower to spawn lefthook through `npx`. Also, removing this assumes dependency on npx, yarn, etc.

PikachuEXE: I re-read #188 and now I understand this is required

Issue #241, opened 2021-10-11
`yarn` conditional may be unreachable in hook template.


andrewhampton:
The [hook template](https://github.com/evilmartians/lefthook/blob/master/cmd/templates/hook.tmpl#L24-L29) currently contains this logic:

```shell
  elif npx @arkweid/lefthook -h >/dev/null 2>&1
  then
    npx @arkweid/lefthook $@
  elif yarn lefthook -h >/dev/null 2>&1
  then
    yarn lefthook $@
```

However, I'm not sure it's possible for the yarn branch to ever execute. I think, if yarn is installed, npx will almost certainly be installed. If `npx` is available on the machine, it will automatically download and run `@arkweid/lefthook`.

I think these branches should be swapped. The yarn branch should only work if lefthook is already installed locally. If it's not already available locally, falling back to npx seems like a good option.

Alternatively, npx has a `--no-install` flag that will prevent it from automatically installing lefthook.

dannobytes:
Got a PR up with the suggested change, hopefully this helps.

mrexox:
Now in master, going to release it in v0.8.0

PR #242, opened 2021-10-11
Fix unreachable conditional in hook template

dannobytes:
Resolves https://github.com/evilmartians/lefthook/issues/241

For devs who are using yarn as their package manager, the git hook
templates are conditioned so that it will never attempt to run `yarn
lefthook` from the local package it is installed in. Instead, the
conditional will always end with `npx @arkweid/lefthook` which ends up
fetching the package via network.

In the worst case, the dev will fetch this package twice, once in the
`elif` conditional and again within the `then` block, causing a
significiant delay before the lefthook binary is executed. This degrades
performance to an almost unusable state.

To fix, give `yarn` a chance to find the lefthook binary within its
local cache before moving on to `npx`.

### DIFF
--- a/cmd/templates/hook.tmpl
+++ b/cmd/templates/hook.tmpl
@@ -21,12 +21,12 @@ call_lefthook()
   elif bundle exec lefthook -h >/dev/null 2>&1
   then
     bundle exec lefthook $@
-  elif npx @arkweid/lefthook -h >/dev/null 2>&1
-  then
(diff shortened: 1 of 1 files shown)


andrewhampton:
Thanks for the reply @Envek. I think the key point here is that the `yarn` branch will never execute as currently written.  `npx` will download lefthook if it's not currently installed.

mrexox:
Great catch! Sorry for the late review.

PR #254, opened 2022-01-13
Refactor install, uninstall, and add commands 

mrexox:
Refactor `install`, `uninstall`, and `add` commands for the lefthook.

Decided to get the logic of lefthook out of cobra configuration. This allows to test the main logic better and also allows to switch CLI-tool to something more lightweight.

- [x] Implement `install` command
- [x] Implement `uninstall` command
- [x] Implement `add` command
- [x] Add tests

Tried to move flags like Force or Aggressive from the global scope to local scopes of install and uninstall commands. Now we support both variants.


pkg/cmd/add.go
```diff
@@ -2,47 +2,44 @@ package cmd
...
-	addCmd := &cobra.Command{
-		Use:   "add",
+	addCmd := cobra.Command{
+		Use:   "add hook-name",
 		Short: "This command add a hook directory to a repository",
-		Long:  longAddCmd,
-		Args:  cobra.MinimumNArgs(1),
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return addExecutor(args, appFs)
+		Long: heredoc.Doc(`
```
palkan, line 16: What about using the `embed` package and put the long description into a text file? https://pkg.go.dev/embed

mrexox:
Closing this in favor of https://github.com/evilmartians/lefthook/pull/264. That PR is rebased on this one.

PR #261, opened 2022-02-27
Run command implementation

markovichecha:


internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+		}
+	}
+	files = filesEsc
+
+	log.Debug("Files after escaping: \n", files)
+
+	runner = strings.Replace(runner, pushFiles, strings.Join(files, " "), -1)
+	runner = strings.Replace(runner, subStagedFiles, strings.Join(files, " "), -1)
+	runner = strings.Replace(runner, subAllFiles, strings.Join(files, " "), -1)
+	runner = strings.Replace(runner, subFiles, strings.Join(files, " "), -1)
```
DarthSim, line 182: We do the same `strings.Join` multiple times. DRY please

DarthSim: Also, it doesn't look right to me that we replace different subs with the same value. @mrexox please check

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+
+	if strings.Contains(runner, subStagedFiles) {
+		files, _ = l.repo.StagedFiles()
+	} else if strings.Contains(runner, subFiles) || filesCommand != "" {
+		files, _ = git.FilesByCommand(filesCommand)
+	} else if strings.Contains(runner, pushFiles) {
+		files, _ = l.repo.PushFiles()
+	} else {
+		files, _ = l.repo.AllFiles()
+	}
```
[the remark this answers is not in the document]
DarthSim: It looks like we should craft different sets of files for each sub and use them for replacement below.

internal/lefthook/run.go
DarthSim: BTW `strings.ReplaceAll` is a bit shorter and more readable

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+
+func setPipeBroken() {
+	isPipeBroken = true
+}
+
+func isExecutable(executable os.FileInfo) error {
+	mode := executable.Mode()
+
+	if !mode.IsRegular() {
+		return errors.New("ErrPermission")
```
DarthSim, line 308: I'd reuse os.ErrPermission here

internal/lefthook/run_test.go
```diff
@@ -0,0 +1,168 @@
...
+			if tt.isErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			require.ElementsMatch(t, tt.okList, okList)
+			require.ElementsMatch(t, tt.failList, failList)
+		})
+		dropState()
```
DarthSim, line 130: I'd make this deferred inside of `t.Run`

internal/lefthook/run_test.go
```diff
@@ -0,0 +1,168 @@
...
+	require.NoError(t, err)
+	lefthook := &Lefthook{Options: &Options{Fs: fs}, repo: repo}
+
+	err = lefthook.Run("post-commit", nil)
+	require.Error(t, err)
+
+	require.ElementsMatch(t, nil, okList)
+	require.ElementsMatch(t, []string{"test0"}, failList)
+	require.Equal(t, true, isPipeBroken)
+	dropState()
```
DarthSim, line 161: This should be deferred too

DarthSim (changes requested):

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+	sourcePath := filepath.Join(sourceDir, hookName)
+	executables, err := afero.ReadDir(l.Fs, sourcePath)
+	if err != nil || len(executables) == 0 {
+		return
+	}
+
+	for _, executable := range executables {
+		wg.Add(1)
+
+		f := func() {
```
mrexox, line 223: There is a race here. If we use parallel running with, say, 3 scripts, all three runs are going to run just one script.

```
.lefthook/
└── pre-commit
    ├── run.go
    ├── run.js
    └── run.sh
```
lefthook.yml:
```

pre-commit:
  parallel: true
  commands:
    1_echo:
      run: echo 1
    2_echo:
      run: echo 2
  scripts:
    "run.go":
      runner: go run
    "run.sh":
      runner: bash
    "run.js":
      runner: node
```

Calling:
```
$ ./lefthook run pre-commit
Lefthook v1.0.0
RUNNING HOOK: pre-commit

  EXECUTE > 2_echo
2


  EXECUTE > 1_echo
1


  EXECUTE > run.sh
shell


  EXECUTE > run.sh
shell


  EXECUTE > run.sh
shell


SUMMARY: (done in 1.01 seconds)
✔️  2_echo
✔️  1_echo
✔️  run.sh
✔️  run.sh
✔️  run.sh
```

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+	} else {
+		spinner.RestartWithMsg(stageName, commandOutput.String())
+
+		failList = append(failList, commandName)
+		setPipeBroken()
+	}
+}
+
+func (l *Lefthook) executeScripts(hookName string, gitArgs []string, sourceDir string, hook *config.Hook, wg *sync.WaitGroup) {
+	sourcePath := filepath.Join(sourceDir, hookName)
```
mrexox, line 214: This is the only line where we need hookName and sourceDir. We can call `filepath.Join` on the caller side and pass only sourcePath here, I think it would be more readable.

internal/lefthook/run.go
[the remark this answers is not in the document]
mrexox: In current implementation `runner` can have only one `{some_files}` template. And it would cause a bug, if we use a few templates here. Although it is hard to imagine a case when we use, say, `{push_files}` and `{staged_files}` in one command, it's better to handle this case properly.

internal/lefthook/run_command.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	return RunPlainCommand(command)
+}
+
+func RunPlainCommand(command *exec.Cmd) (*bytes.Buffer, WaitFunc, error) {
+	ptyOut, err := pty.Start(command)
+
+	// Copy stdin to the pty and the pty to stdout.
+	go func() { _, _ = io.Copy(ptyOut, os.Stdin) }()
```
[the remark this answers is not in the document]
mrexox: Right, we don't need it. This is a line from README of the package btw, I don't think there was another reason to use it in current implementation.

internal/lefthook/run.go
mrexox: We assume that we have only one template in the string, that's why it worked. According to the previous discussion, let's support multiple templates in the `run` string?

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+				runnerArg := append(strings.Split(runner, " "), executablePath)
+				runnerArg = append(runnerArg, gitArgs...)
+
+				command = exec.Command(runnerArg[0], runnerArg[1:]...)
+			}
+
+			commandOutput, wait, err := RunPlainCommand(command)
+
+			mutex.Lock()
+			defer mutex.Unlock()
```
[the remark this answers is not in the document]
markovichecha: What about modifying fail and ok lists?

internal/lefthook/run.go
```diff
@@ -10,6 +56,293 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+		return err
+	}
+	if err := cfg.Validate(); err != nil {
+		log.Info(log.Brown(fmt.Sprintf("Config error! %s", err)))
+		return err
+	}
+
+	if tags := os.Getenv("LEFTHOOK_EXCLUDE"); tags != "" {
+		envExcludedTags = strings.Split(tags, ",")
+	}
```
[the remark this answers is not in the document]
markovichecha: :+1:

internal/lefthook/run.go
markovichecha: yeah, obvious error :-1:

internal/lefthook/run.go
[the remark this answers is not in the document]
DarthSim: WDYT about wrapping it to functions with their own mutex? We won't lock the whole thing this way.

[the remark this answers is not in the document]
mrexox: Or we can use channels for fail and ok lists, or even one channel for all results. Of course, if it seems reasonable here.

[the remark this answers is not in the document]
DarthSim: Nah, I think channels will bring more headache than use. We'll need to ensure they're buffered enough, etc.

internal/lefthook/run_command.go
```diff
@@ -0,0 +1,38 @@
...
+	commandOutput := bytes.NewBuffer(make([]byte, 0))
+	_, _ = io.Copy(commandOutput, ptyOut)
+	waitFunc := func() error {
+		wErr := command.Wait()
+		_ = ptyOut.Close()
+		return wErr
+	}
```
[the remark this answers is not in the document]
markovichecha: We use it for making sure that we check if it's execution error. I'll add bool parameter showing that command execution started.

internal/lefthook/run.go
```diff
@@ -10,6 +49,313 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+			runner = strings.ReplaceAll(runner, config.PushFiles, strings.Join(files, " "))
+		}
+	}
+	if strings.Contains(runner, config.SubAllFiles) {
+		files, err := l.repo.AllFiles()
+		files = prepareFiles(command, files)
+		if err == nil {
+			runner = strings.ReplaceAll(runner, config.SubAllFiles, strings.Join(files, " "))
+		}
+	}
```
DarthSim, line 168: Can we DRY this?

internal/lefthook/run_command.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func RunPlainCommand(command *exec.Cmd) (*bytes.Buffer, bool, error) {
+	ptyOut, err := pty.Start(command)
+	if err != nil {
+		return nil, false, err
+	}
+	defer ptyOut.Close()
+
+	commandOutput := bytes.NewBuffer(make([]byte, 0))
```
DarthSim, line 33: Not critical, yet `new(bytes.Buffer)` does the same thing as `bytes.NewBuffer(make([]byte, 0))`

internal/lefthook/run_command.go
```diff
@@ -0,0 +1,40 @@
...
+	_, err = io.Copy(commandOutput, ptyOut)
+	if err != nil {
+		log.Debug(err)
+	}
```
DarthSim, line 37: This should probably be a goroutine as the copying won't stop until `ptyOut` is closed

DarthSim (changes requested):
Sorry, I promised there will be a single review, but there are a couple of things I'd like to change

internal/lefthook/run.go
```diff
@@ -10,6 +49,313 @@ func Run(opts *Options, hookName string, gitArgs []string) error {
...
+		log.Info(log.Cyan(fmt.Sprintf("\nSUMMARY: (done in %.2f seconds)", execTime.Seconds())))
+	}
+
+	for _, fileName := range okList {
+		log.Infof("✔️  %s\n", log.Green(fileName))
+	}
+
+	for _, fileName := range failList {
+		log.Infof("🥊  %s", log.Red(fileName))
+	}
```
mrexox, line 360: Missing "\n"

mrexox (commented):
Also added a few minor comments. 
Could you, please, fix linter complains and tests?

internal/lefthook/run.go
markovichecha: Do you have any suggestions? I can't find any obvious solutions.

internal/lefthook/run_command.go
markovichecha: Copying will stop when reader receive EOF. Do you think there might by some problems with that?

internal/lefthook/run.go
mrexox: What about creating an array of anonymous structures, that hold the dependencies of each part? (the substitution template and a git function that returns the files and error)

markovichecha: Yeah, I thought about that, but one of the functions have a different signature.

DarthSim: I think about something like this:

```go
func replaceTemplate(runner, sub string, f func()([]string, error)) string {
  if !strings.Contains(runner, config.SubStagedFiles) {
    return runner
  }
  
  files, err := f()
  // ...
}
```

markovichecha: I'll just wrap it

internal/lefthook/run_command.go
DarthSim: AFAIK `ptyOut` won't return EOF unless it is closed. Not sure though. Gorouitine won't hurt anyone anyway I believe.

markovichecha: It will, actually. Because we modify a commandOutput and also return it in main flow.

DarthSim: Damn, yeah, you're right. On the other hand, we wait for the command completion and close the pipe on exit, so `commandOutput` shouldn't be modified after this function. Yet it's a pretty thin edge.

Also, I found this https://github.com/creack/pty/issues/127. It seems like tty is closed too early, so there may be problems with long-running processes.

However, I checked the code in master and it looks like everything should work like this

mrexox:
Amazing work! I am going to merge this PR and continue here: https://github.com/evilmartians/lefthook/pull/264. There are some tests that need fixing for windows runner, and goreleaser config that doesn't know how to build the app yet. This is not a problem this PR should solve.

As you see we are refactoring the current codebase step by step in the `refactoring` branch. When everything is ready we will release the new version of lefthook. If this project is interesting for you and you have time and wish, please come with feature requests or any other questions. It would be great to work on this.

Anyway, thank you for the contribution!

markovichecha:
> Amazing work! I am going to merge this PR and continue here: https://github.com/evilmartians/lefthook/pull/264. There are some tests that need fixing for windows runner, and goreleaser config that doesn't know how to build the app yet. This is not a problem this PR should solve.
> 
> As you see we are refactoring the current codebase step by step in the `refactoring` branch. When everything is ready we will release the new version of lefthook. If this project is interesting for you and you have time and wish, please come with feature requests or any other questions. It would be great to work on this.
> 
> Anyway, thank you for the contribution!

Thanks, it was a fun experience for me :)

Issue #279, opened 2022-06-21
Skip command in `rebase` not working in interactive rebases


p4checo:
I've recently upgraded to version 1.0.0 and enabled skip functionality on merges and rebases for my pre-commit hook, but the hook is still invoked when I do interactive rebases.

My `lefthook.yml`:
```yml
pre-commit:
  parallel: true
  commands:
    swiftlint:
      run: |
        swiftlint --fix;
        (cd FolderA && swiftlint --fix);
        (cd FolderB && swiftlint --fix)
      skip:
        - merge
        - rebase
```

however when I do a `git rebase -i` I still get messages like these:

```
> git rebase -i HEAD~1
Lefthook v1.0.0
RUNNING HOOK: pre-commit
⠙ waiting
```

And I also get messages like this in between rebase commits, despite not having any `prepare-commit-msg` hook configured in `lefthook.yml`...

```
Lefthook v1.0.0
RUNNING HOOK: prepare-commit-msg
```

Can you please help? Thanks! 🙏🏼

mrexox:
Hi! Thank you for the issue! Could you clarify some details, please :pray:

- Does `pre-commit` hook command (`swiftlint`) gets executed on `git rebase`, or does it hang forever? Could you provide the full output if it gets executed?

- Historically we used `prepare-commit-msg` just to store lefthook config checksum. We were skipping this hook. With version `1.0.0` we still keep checksum in `prepare-commit-msg` but also allow to configure this hook. I think we can fix it with the next release and store the checksum somewhere else. Is it just annoying or breaks things?

p4checo:
Hi @mrexox! 👋🏼 

Thanks for the quick reply!

> - Does pre-commit hook command (`swiftlint`) gets executed on `git rebase`, or does it hang forever? Could you provide the full output if it gets executed?

if I execute a normal `git rebase` then the hook (`swiftlint`) isn't executed, only the `prepare-commit-msg` for each commit in the branch (I had 3 commits)

```
> git rebase test-rebase                                                                                                                                                                                   
Lefthook v1.0.0
RUNNING HOOK: prepare-commit-msg
Lefthook v1.0.0
RUNNING HOOK: prepare-commit-msg
Lefthook v1.0.0
RUNNING HOOK: prepare-commit-msg
Successfully rebased and updated refs/heads/test-skip.
```

[...]

PR #280, opened 2022-06-21
Store lefthook checksum in non-hook file

mrexox:
Relates to https://github.com/evilmartians/lefthook/issues/279

- Store lefthook config checksum not in `prepare-commit-msg`, but in `lefthook.checksum` file.
- Make sure config is synchronized when calling `run` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
+COMMIT_HASH = $(shell git rev-parse HEAD)
+
 build:
-	go build -o lefthook
+	go build -ldflags "-X github.com/evilmartians/lefthook/internal/version.commit=$(COMMIT_HASH)" -o lefthook
 
 test:
 	go test -cpu 24 -race -count=1 -timeout=30s ./...
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -9,13 +9,20 @@ import (
 )
 
 func newVersionCmd(opts *lefthook.Options) *cobra.Command {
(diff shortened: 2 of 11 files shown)


Envek:
Is it a good idea to still store checksums in the hooks directory? It is not a hook now.

Also, as far as I remember, purpose of `prepare-commit-msg` hook was to detect config changes and install or uninstall hooks automatically. How it will be done now?

mrexox:
> Is it a good idea to still store checksums in the hooks directory? It is not a hook now.
> 
> Maybe we should store it in [`.git/info`](https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-info) instead? 
> 
> Also, as far as I remember, purpose of `prepare-commit-msg` hook was to detect config changes and install or uninstall hooks automatically. How it will be done now?

@Envek Thank you! `.git/info` is much better directory for this purpose.

Well, while refactoring I have forgotten about this feature :confused:. I have an idea how bring it back. We could store config file *last updated timestamp* with the checksum and compare it on every `run` command. It is cheaper that calculating checksum on every call.

If we find that the timestamp has changed, we can invoke `install` command, which compares the checksums and installs if necessary.

This might be **not a very clean** solution but I think it is more fair to let `prepare-commit-msg` hook to be just a usual configurable hook. (I missed this feature when I was using 0.* versions). 

What do you think?

PR #282, opened 2022-06-22
chore: fix version

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/104130

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -6,7 +6,7 @@ import (
 	"strconv"
 )
 
-const Version = "1.0.0"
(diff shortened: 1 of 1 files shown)


mrexox:
Hey! I am sorry, the binary inside v1.0.1 tag still contains v1.0.0 version. But I am going to release 1.0.2 soon. Let's leave 1.0.1 as a version that is missed on homebrew (actually it doesn't introduce anything new, just a fix for NPM and Rubygems). I hope to never create such a mess. :pray:

Issue #296, opened 2022-07-14
Lefthook crash - `Unknown OS: mingw (RuntimeError)`


ariccio:
On a first time fresh install of lefthook, it crashed on `lefthook --help`:
```
lefthook --help
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/lefthook-1.0.4/bin/lefthook:21:in `<top (required)>': Unknown OS: mingw (RuntimeError)
        from C:/Ruby31-x64/bin/lefthook:32:in `load'
        from C:/Ruby31-x64/bin/lefthook:32:in `<main>'
```

Relevant bits of `bundle env`:

```
Bundler       2.3.10
  Platforms   ruby, x64-mingw-ucrt
Ruby          3.1.1p18 (2022-02-18 revision 53f5fc4236a754ddf94b20dbb70ab63bd5109b18) [x64-mingw-ucrt]
  Full Path   C:/Ruby31-x64/bin/ruby.exe
```

I suspect it has to do with all the native library changes that have been going on in Ruby. Not that anybody asked, but however much I love ruby, I will never understand why they chose to use an abomination like mingw. *shakes fist at unix compat layers* :)

Related issues:

- https://github.com/evilmartians/lefthook/issues/278
- https://github.com/evilmartians/lefthook/issues/79
- https://github.com/rubygems/rubygems/issues/5269

PR #297, opened 2022-07-14
Add `when "mingw" then "windows"` case

ariccio:
Making this change locally allowed me to run lefthook. I do not *know* if there are any kind of ABI issues, I have no *actual* idea how lefthook works, but I suspect it's fine since it looks like there's no actual interface between go and ruby, only at the process level?

Fixes https://github.com/evilmartians/lefthook/issues/296

### DIFF
--- a/packaging/rubygems/bin/lefthook
+++ b/packaging/rubygems/bin/lefthook
@@ -18,6 +18,7 @@ os =
   when "darwin"  then "darwin"  # MacOS
   when "windows" then "windows"
   when "mingw32" then "windows" # Windows with MINGW64 reports RUBY_PLATFORM as "x64-mingw32"
+  when "mingw" then "windows"
   else raise "Unknown OS: #{platform.os}"
   end
 


mrexox:
Hey! Thank you very much for the fix!

ariccio:
Heh, of all the weird environment issues I've faced the past few months, this was one of the only ones that was easy to figure out and fix!

Among them:
1. Expo failing to connect to my laptop because for some reason only two devices on my local network have bad ARP tables, those two are my android dev device and my laptop, and the only thing missing from the table was precisely the two devices I wanted to talk to eachother... The only way I fixed that was `nmap`ing the network, which apparently caused *someone* to update.
2. Windows deciding to *reserve* ports in ways that other platforms seemingly don't, which leads to [all](https://github.com/puma/puma/issues/2894) [kinds](https://github.com/strapi/strapi/issues/5057) [of](https://stackoverflow.com/q/62508193/625687) [confusion](https://superuser.com/q/1560888).
3. [`libuv` in `node` not handling errors in a generic way, so uncommon OS errors (like subtle disk corruption) cause hard-to-debug crashes.](https://github.com/nodejs/node/issues/43439) That one took a while.

mrexox:
Hey! I have released your fix with v1.0.5. Please, come back if you have any problem!

Issue #294, opened 2022-07-10
`lefthook` package is missing `postinstall ` script


fantua:
Trying to use `lefthook` package according to this [doc](https://github.com/evilmartians/lefthook/blob/master/docs/node.md#install) via:
```console
yarn add -D lefthook
```

But pre-commit hooks doesn't work because there is no `postinstall` script:
https://github.com/evilmartians/lefthook/blob/abf62bc11d95b96c82ad95797b6eda56aaf6bc08/packaging/npm/lefthook/package.json#L1-L30

It exist only for `@evilmartians/lefthook` package:
https://github.com/evilmartians/lefthook/blob/abf62bc11d95b96c82ad95797b6eda56aaf6bc08/packaging/npm-bundled/package.json#L34

Is it expected?

mrexox:
Hi! It is not expected because lefthook package has optional dependencies which resolve for each OS and arch.

mrexox:
@fantua Did you run `yarn run lefthook install` after installation?

fantua:
> @fantua Did you run yarn run lefthook install after installation?

No, I didn't. I expected this to be done automatically (using `postinstall` hook). Same as it’s done in `@evilmartians/lefthook`

PR #310, opened 2022-08-07
Add install command to npm package

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/294

**Summary :zap:**

- [x] Add install.js script to all os-arch dependencies

### DIFF
--- a/packaging/npm/lefthook-darwin-arm64/install.js
+++ b/packaging/npm/lefthook-darwin-arm64/install.js
@@ -1,3 +1,4 @@
+const path = require("path")
 const { spawnSync } = require("child_process")
 
 function install() {
--- a/packaging/npm/lefthook-darwin-arm64/install.js
+++ b/packaging/npm/lefthook-darwin-arm64/install.js
@@ -17,5 +17,5 @@ function install() {
 try {
   install()
 } catch(e) {
-  console.warn("'lefthook install' command failed. Try running it manually")
+  console.warn("'lefthook install' command failed. Try running it manually.\n" + e)
 }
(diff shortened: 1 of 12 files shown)


fantua:
Is it possible that someone will use `lefthook-*` package directly without `lefthook`? 
If not, than maybe we can just add one `install` script to [lefthook](https://github.com/mrexox/lefthook/blob/704b0dcafa9a2cee77980f93de27df34b383e81d/packaging/npm/lefthook/package.json) package directly? Something like:
```js
const { spawnSync } = require("child_process")
const { getExePath } = require('./get-exe');

function install() {
  if (process.env.CI) {
    return
  }

  spawnSync(getExePath(), ["install", "-f"], {
    cwd: process.env.INIT_CWD || process.cwd(),
    stdio: "inherit",
  })
}

try {
  install()
} catch(e) {
  console.warn("'lefthook install' command failed. Try running it manually.\n" + e)
}
```
In `install` phase all dependencies should be already installed.

Also Yarn 3 (berry) [supports](https://yarnpkg.com/advanced/lifecycle-scripts) `install` script only for backwards compatibility. Maybe it's better to use `postinstall` instead (supported by both npm and yarn) 🙂

mrexox:
Nice suggestion, @fantua! I read about [issues](https://github.com/npm/npm/issues/4134) with NPM installing dependencies before `postinstall` scripts, but I guess this was fixed for the last 3-5 years.

The fix: https://github.com/evilmartians/lefthook/pull/311

Btw, feel free to open PRs :)

Issue #269, opened 2022-05-05
Bad quote escape with "git-bash" (windows)


TotoLastro:
Hi all,

After installing lefthook from npm, I did some tests.
And I see a problem with quotes (a bad escape ?).

For the example, this is the lefthook.yml :
```yaml
pre-commit:
  parallel: false
  commands:
    echo-files:
      run: echo "{staged_files}"
```
And this is the result after running the command :
```
$ lefthook -v run pre-commit
RUNNING HOOKS GROUP: pre-commit
  waiting Files before filters:
 [lefthook.yml]
Files after filters:
 [lefthook.yml]
Files after escaping:
 [lefthook.yml]

  EXECUTE > echo-files
\lefthook.yml"


SUMMARY: (done in 0.17 seconds)
```
(we can see trailing chars : **\\** and **"**)

No problem if I use apostrophes (the **echo** command prints only the filename)

- lefthook 0.7.7
- Git Bash (x64) 2.36.0.1
- Windows 10 Enterprise N 20H2

mrexox:
Hey! We've released new version of lefthook recently. Could you check if there's still a bug with the latest lefthook binary?

Carduelis:
@mrexox Unfortunately, it doesn't work for me.
I'm using next.js and there is a common pattern to have the `[slug].tsx` filenames.
I'm on Windows and if I run the command with double quotes, it works:
```
yarn eslint "pages/blog/[slug].tsx"
```

But the lefthook uses single quotes which leads to the following error:
```
Oops! Something went wrong! :(

ESLint: 8.17.0

No files matching the pattern "'pages/blog/[slug].tsx'" were found.
Please check for typing mistakes in the pattern.
```

Carduelis:
My `lefthook.yml` is:
```
pre-commit:
  parallel: true
  commands:
    prettier:
      glob: '*.{js,ts,jsx,tsx}'
      run: yarn prettier --write {staged_files} && git add {staged_files}
    eslint:
      glob: '*.{js,ts,jsx,tsx}'
      run: yarn eslint {staged_files} --fix && git add {staged_files}
```

How can I replace `'` with `"`?

Carduelis:
For those, who experience issues using next.js framework with [Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes), there is a temporal solution to ignoring files for optional wildcard routing.

[...]

PR #314, opened 2022-08-10
Windows escaping issues

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/269

**Summary :wrench:**

This bug is from go `exec` builtin package: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/os/exec/exec.go;l=269. They suggest passing args directly with syscall attrs. This is what is done with this PR.

Also fixed a thing: if the {staged_files} or any other template is quoted like "{staged_files}", all the files from substitution will be quoted. Previously the whole line was quoted, and this is not what you want on Windows, I think.

### DIFF
--- a/docs/full_guide.md
+++ b/docs/full_guide.md
@@ -168,6 +168,24 @@ pre-commit:
 
 Note: If using `all_files` with RuboCop, it will ignore RuboCop's `Exclude` configuration setting. To avoid this, pass `--force-exclusion`.
 
+If you want to have all you files quoted with double quotes `"` or single quotes `'`, quote the appropriate shorthand:
+
+```yml
+pre-commit
+  commands:
+    lint:
+      glob: "*.js"
+      # Quoting with double quotes `"` might be helpful for Windows users
+      run: yarn eslint "{staged_files}" # will run `yarn eslint "file1.js" "file2.js" "[strange name].js"`
+    test:
+      glob: "*.{spec.js}"
+      run: yarn test '{staged_files}' # will run `yarn eslint 'file1.spec.js' 'file2.spec.js' '[strange name].spec.js'`
+    format:
+      glob: "*.js"
+      # Will quote where needed with single quotes
+      run: yarn test {staged_files} # will run `yarn eslint file1.js file2.js '[strange name].spec.js'`
+```
+
 ## Custom file list
 
 Lefthook can be even more specific in selecting files.
--- a/internal/lefthook/runner/execute_windows.go
+++ b/internal/lefthook/runner/execute_windows.go
@@ -5,12 +5,18 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
+	"syscall"
 )
 
 type CommandExecutor struct{}
 
 func (e CommandExecutor) Execute(root string, args []string) (*bytes.Buffer, error) {
-	command := exec.Command(args[0], args[1:]...)
+	command := exec.Command(args[0])
+	command.SysProcAttr = &syscall.SysProcAttr{
+		CmdLine: strings.Join(args, " "),
+	}
+
 	rootDir, _ := filepath.Abs(root)
 	command.Dir = rootDir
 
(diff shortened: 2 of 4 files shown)


ariccio:
Thanks for paying attention to us windows users :)

But it looks like this patch inadvertently added a regression. I'm not 100% sure what the offending change was in this commit.

I have some scripts declared in my `lefthook.yml` as such:

```
pre-push:
  parallel: true
  scripts:
    "check_validates_timeliness.rb":
      runner: ruby
    "check_if_googleauth_compat_with_heroku_22.rb":
      runner: ruby

```

These scripts are in a path with a space in a parent directory's name - my username! - and it looks like lefthook is no longer passing the path directly to invoke the new process. So ruby gets invoked with a path that ends at the space. lefthook eventually spits out an error like:

```
⠼ waitingC:/Users/Lucius:1: numeric literal without digits
[0B48:3374][2022-08-09T17:26:40]...
C:/Users/Lucius:1: syntax error, unexpected integer literal, expecting ']'
[0B48:3374][2022-08-09T17:26:40]i0...
C:/Users/Lucius:1: numeric literal without digits
[0B48:3374][2022-08-09T17:26:40]...
C:/Users/Lucius:1: syntax error, unexpected integer literal, expecting ']'
[0B48:3374][2022-08-09T17:26:40]i0...

  EXECUTE > check_validates_timeliness.rb


  EXECUTE > check_if_googleauth_compat_with_heroku_22.rb


SUMMARY: (done in 2.18 seconds)
🥊  check_validates_timeliness.rb
🥊  check_if_googleauth_compat_with_heroku_22.rb

```

(yes, lmao, this was my dad's computer like 10 years ago, I never changed the username)

...as seen in `procmon`, `lefthook` invokes ruby with the following unquoted command line:
```
ruby C:\Users\Lucius Riccio\Documents\GitHub\COVID-CO2-tracker\.lefthook\pre-push\check_validates_timeliness.rb
```

*mimes shaking fist* path names are a such a pain! :D

mrexox:
Hey! Thank you for the issue description! I'll check this. This is definitely fixable!

mrexox:
Hay @ariccio , can you update to 1.1.1? This issue should be fixed there

ariccio:
Ok, I couldn't get to it today, and I'm working the polls tomorrow (all day! 5AM - 11pm!), But I'll try and get to it Wednesday!

ariccio:
Uh, I may have been stuck distracted by my family for like a week after working the polls (sshhh, don't tell them). I've updated lefthook, and am now seeing a different problem... looks [like `chmod`ing is failing](https://github.com/mrexox/lefthook/blob/9914c98e8e178b6aa8fe677a0fae7a9c20ce6f1e/internal/lefthook/runner/runner.go#L188).

Looks like in go on Windows, `chmod` just calls into [SetFileAttributesW](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileattributesw), nothing fancy going on. The path in the error message is surrounded by single quotes. Is it possible that you're now passing the quoted path to the Win32 api? That's my guess here. Honestly, it's kind of amazing we're still using text strings/command lines to do anything these days! APIs that don't need quotes are the way to go :)

mrexox:
Oh, I'll take a look at that issue. This is really surprising, because it's Go API, that **should** be working :thinking:.

ariccio:
I'm not a go programmer, I don't think I can easily hotpatch this locally right? I suspect it's not a problem with go, but that windows doesn't like quoted paths being passed to "strongly typed" APIs (which don't need no stinkin quotes)... I think you might have been passing the newly quoted path through to the other part of the program?

ariccio:
I'll tell ya, as one of the few programmers out there who has made the mistake of a main user account with a space in it, nothing else has caused more pain, or uncovered more bugs, in my entire career 🤣

ariccio:
@mrexox, I submitted a PR to fix the regression.

Issue #320, opened 2022-08-21
Pass local and remote branch names as arguments / env variables to pre-push


JoyceBabu:
Is it possible to access the local / remote ref names in the `pre-push` hook?

Git passes the information as STDIN to the `pre-push` hook. But lefthook is not passing it to the actual hook.

mrexox:
Yes, lefthook doesn't pass STDIN now. Could you describe your case?

JoyceBabu:
I am using a pre-push hook to statically analyze the source code using PHPStan.

Since it is a legacy code base, the code is not fully typehinted, and there are several thousand existing issues.

We want to ensure that new changes does not introduce additional issues, while we are are slowly fixing the existing errors.
For this, we have a `pre-push` hook that statically analyzes the codebase twice. First for the merge base to create a baseline of existing issues, and then for the pushed commit. All issues that are missing in the baseline, are expected to be fixed, before the branch is pushed.

To determine the merge base, we need access to the local ref name.

mrexox:
Wow! This looks reasonable. Could you share your `lefthook.yml`? I think passing STDIN is possible but I need to find the most proper way to do so.

Does PHPStan support analyzing only the given files? I mean, you could run it only for files that were touched in commits that are being pushed. There is a `{push_files}` template for that. So, it would look like this:

```yaml
pre-push:
  commands:
    phpstan:
      run: phpstan {push_files}
```

If it doesn't help, and using STDIN is the only option, I would ask you to provide some reproducible code+lefthook.yml set, so I could test STDIN passing while developing a feature.

JoyceBabu:
I am not sure whether static analysis is possible with just a subset of files. Changes in a file can affect code in unmodified files too.

Also, the hook is supposed to report only new issues. Because, sometimes the modifieds may have hundreds of issues and it may not be  feasible to fix them all in a single go.

There is an [existing issue](https://github.com/evilmartians/lefthook/issues/147) about the missing STDIN.

I have created a test repo with the static analysis check
https://github.com/JoyceBabu/lefthook-test

I had to [hard code](https://github.com/JoyceBabu/lefthook-test/blob/master/phpstan-commit#L55) the values for `local_ref` as `HEAD` and `remote_ref` as `origin/master`.

mrexox:
Thank you a lot! I will start working on it soon! (wow, you made a repo for this case! I appreciate it a lot!)

mrexox:
I thought a little bit and came up with the following solution:

[...]

PR #324, opened 2022-08-29
Pass stdin by default

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/320

**Summary :wrench:**

- [x] Pass STDIN
- [x] Add `interactive` boolean option instead of always using `exec < /dev/tty` in hook template


mrexox:
Will be released in 1.2.0 in 1-2 weeks.

PR #336, opened 2022-10-13
Print command name before execution

abitrolly:
To help read the cause when `lefthook` hangs indefinitely.

Tested. With `parallel: true` results are very messy.

### DIFF
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -397,13 +397,13 @@ func replaceQuoted(source, substitution string, files []string) string {
 }
 
 func (r *Runner) run(opts RunOptions) {
+	var execName string
 	if !r.logSettings.SkipExecution() {
 		execName = fmt.Sprint(log.Cyan("\n  EXECUTE >"), log.Bold(opts.name))
 		log.Infof("%s\n", execName)
 	}
 	out, err := r.exec.Execute(opts.root, opts.args, opts.interactive)
 
-	var execName string
 	if err != nil {
 		r.fail(opts.name, opts.failText)
 		execName = fmt.Sprint(log.Red("\n  EXECUTE >"), log.Bold(opts.name))
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -397,6 +397,10 @@ func replaceQuoted(source, substitution string, files []string) string {
 }
 
 func (r *Runner) run(opts RunOptions) {
+	if !r.logSettings.SkipExecution() {
+		execName = fmt.Sprint(log.Cyan("\n  EXECUTE >"), log.Bold(opts.name))
+		log.Infof("%s\n", execName)
+	}
 	out, err := r.exec.Execute(opts.root, opts.args, opts.interactive)
 
 	var execName string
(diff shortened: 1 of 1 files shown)


mrexox:
Hey! Let me help you investigate the problems with hangs. Could you provide your **lefthook.yml** that hangs? Maybe I can help you solve this problem.

As you noticed, with `parallel: true` results can be very messy. I believe we can solve this problem the other way.

abitrolly:
@mrexox here is it https://gitlab.com/gitlab-org/gitlab/-/blob/master/lefthook.yml

mrexox:
Hey @abitrolly, I've tried to experiment on my private cloned repo with all the checks in lefthook.yml, but they didn't hang for me. Could you provide a PR (or MR if it is GitLab) for which lefthook hangs? You can use `LEFTHOOK=0 git push` to disable lefthook execution on git push.

abitrolly:
@mrerox I am using Gitpod to run GitLab. Maybe it doesn't have the same performance, and the jobs just take a lot of time to complete. In any case waiting for more than 5 minutes in a terminal without any glue about what is going on is not a good user experience. I could skip certain tasks, or look into speeding them up if they are a bottleneck, but I can not identify them.

mrexox:
Hey! I prepared a possible solution: https://github.com/evilmartians/lefthook/pull/338. It will print the current running commands/scripts right after the `waiting` spinner. I think it will be more convenient.

I am going to release it with **1.1.3**. I hope it would help you to find the problem. I also think there might be a bug with lefthook itself. There were hangs on pre-push hooks when the commands/scripts were waiting for the input from STDIN, and I was fixing them in **1.1.2** :thinking:. When I release the new fix, please, come if you find more details on this bug.

mrexox:
Hey @abitrolly! How is it going? Did you have a chance to check **1.1.3**?

abitrolly:
Not yet, but let my try this right now.

abitrolly:
@mrexox works good to show the command name.

```
⠙ waiting: danger, rubocop
```

I synced my branch with upstream, so there are not many changed files, so didn't run it over all tasks over few hours. Let me still try that though.

abitrolly:
Well, I can't see any obvious MRs with a lot of changes to different files. But the next time I encounter them I give it a test. It is still would be good to see timing information by the step in the end.

abitrolly:
@mrexox the terminal width may be determined incorrectly. Observed this on Gitpod.

```
⠸ waiting: stylelint-tests, prettier-tests, yamllint-tests, jest-tests, eslint-tests, markdownlint-tests, r
⠼ waiting: stylelint-tests, prettier-tests, yamllint-tests, jest-tests, eslint-tests, markdownlint-tests, r
⠴ waiting: stylelint-tests, prettier-tests, yamllint-tests, jest-tests, eslint-tests, markdownlint-tests, r
⠦ waiting: stylelint-tests, prettier-tests, yamllint-tests, jest-tests, eslint-tests, markdownlint-tests, r
⠧ waiting: stylelint-tests, prettier-tests, yamllint-tests, jest-tests, eslint-tests, markdownlint-tests, r
```

image

mrexox:
Oh, I see. Isn't it scrollable horizontally?

abitrolly:
Might be Gitpod not reporting correct width of the terminal. Or reducing the width of the terminal after a command already started.

mrexox:
Yes, I think there's something wrong with terminal, because lefthook just prints text, so it is out of its responsibility :thinking:

abitrolly:
@mrexox when printing `lefthook` tries to git fit terminal width. Does it reevaluate the width after it has been already started?

mrexox:
@abitrolly , actually lefthook doesn't have explicit code associated with terminal. Maybe there is something in [spinner](github.com/briandowns/spinner) lib we use :thinking:

abitrolly:
Fixed link https://github.com/briandowns/spinner

PR #338, opened 2022-10-14
Show current running commands and scripts

mrexox:
Closes https://github.com/evilmartians/lefthook/pull/336

**Summary :zap:**

Print the current running commands/scripts:

```yaml
# lefthook.yaml
test:
  parallel: true
  commands:
    command1:
      run: sleep 1
    command2:
      run: sleep 3
    command3:
      run: sleep 4
```

```bash
$ lefthook run test
Lefthook v1.1.2
RUNNING HOOK: test
  EXECUTE > command1

⠧ waiting: command3, command2
```

### DIFF
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -397,6 +397,9 @@ func replaceQuoted(source, substitution string, files []string) string {
 }
 
 func (r *Runner) run(opts RunOptions) {
+	log.SetName(opts.name)
+	defer log.UnsetName(opts.name)
+
 	out, err := r.exec.Execute(opts.root, opts.args, opts.interactive)
 
 	var execName string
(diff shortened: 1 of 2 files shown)


abitrolly:
Nice. If the command runs more than 5 minutes, it may worth to add running time.

I don't know if it is possible to measure bottlenecks for child processes, but if all commands are competing for disk I/O or memory, because they are being run in parallel, it may be worth to warn about that too, to optimize the times needed to run these hooks in future.

mrexox:
Good suggestion. Although it requires some statistics gathering, because we need to understand what is too much time/memory and what is not.

Also limiting the concurrency is a thing to consider. I am welcome for new suggestions, please, come and create some issue :)

Or you can start a discussion instead: https://github.com/evilmartians/lefthook/discussions, and we can continue it there. This is a good point for a discussion, especially for such a big projects like GitLab.

PR #369, opened 2022-11-16
Create Go-syntax representation of the value

stonesbg:
The best I can tell the root of the issue is that the filepath for the script is converted to a single quote and this combined with the exec.Command causes the windows shell to not handle things correct.

This is the only simple change that I could come up with.  It tested with windows and wsl (ubuntu) and both didn't have issues.  The other alternatives that I could think of is rather than combining the args up front and then passing to downstream executioners instead keep the script path as an os.FileInfo type until it gets to the executioners (execute_windows.go & execute_unix).

The last option was to use exec.Command("cmd", "/c", ...) or exec.Command("powershell", "-c", ...) where it just passes n the raw built up command.  Downside is with windows at least you might lose some functionality with cmd/pwsh/powershell based on the shell configuration.

### DIFF
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -222,9 +222,7 @@ func (r *Runner) runScript(script *config.Script, unquotedPath string, file os.F
 		args = strings.Split(script.Runner, " ")
 	}
 
-	quotedScriptPath := fmt.Sprintf("%#v", unquotedPath)
-
-	args = append(args, quotedScriptPath)
+	args = append(args, fmt.Sprintf("%#v", unquotedPath))
 	args = append(args, r.args[:]...)
 
 	if script.Interactive {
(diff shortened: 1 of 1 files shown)


internal/lefthook/runner/runner.go
```diff
@@ -222,7 +222,8 @@ func (r *Runner) runScript(script *config.Script, unquotedPath string, file os.F
 		args = strings.Split(script.Runner, " ")
 	}
 
-	quotedScriptPath := shellescape.Quote(unquotedPath)
+	quotedScriptPath := fmt.Sprintf("%#v", unquotedPath)
```
mrexox, line 225: Wow, so if I correctly understood, with `Sprintf("%#v")` the bug doesn't reproduce, right? Interesting :thinking: 
You know, a few releases before the path for scripts wasn't even quoted. I think this is a regression, and I added the quoting accidentally. 

@stonesbg could you remove [this](https://github.com/stonesbg/lefthook/blob/master/internal/lefthook/runner/runner.go#L225) line and rename the `unquotedPath` argument to be just `path`?

Good job with finding the reason for the buggy behavior! I appreciate it so much!

mrexox: Not at all :) Let me help

mrexox:
Please, check v1.2.1. The issue should be fixed there. Thank you for contributing!

Issue #253, opened 2022-01-11
Command not found on run


tomkys144:
Hi,
when running pre-commit hook (either by `git commit` or by `lefthook run`) I get "command not found" error although in command line and in shell scripts everything works just fine...

config:
```yml
pre-commit:
  piped: true
  commands:
    1_lint_code:
      glob: '*.dart'
      run: dart fix --apply lib && git add .
    2_format_code:
      glob: '*.dart'
      run: flutter format {staged_files} && git add {staged_files}
```

Shell output:
```bash
Lefthook v0.7.7
RUNNING HOOKS GROUP: pre-commit

  EXECUTE > 1_lint_code
 sh: line 1: dart: command not found


 2_format_code (SKIP BY BROKEN PIPE)

SUMMARY: (done in 0.01 seconds)
🥊  1_lint_code
```

And in none-pipe mode:
```bash
Lefthook v0.7.7
RUNNING HOOKS GROUP: pre-commit

  EXECUTE > 1_lint_code
 sh: line 1: dart: command not found


  EXECUTE > 2_format_code
 sh: line 1: flutter: command not found


SUMMARY: (done in 0.02 seconds)
🥊  1_lint_code
🥊  2_format_code
```

mrexox:
Hey! Do you edit your PATH env in .bashrc or something? Or maybe you have an alias? How are `dart` and `flutter` installed on your system?

art-solopov:
I'm having a similar problem. I've aliased `docker-compose` to `docker compose` but it still doesn't work. Does lefthook not respect aliases?

mrexox:
@art-solopov, did you write an alias in `.bashrc` or `.zshrc`? Lefthook uses `sh` as a shell, this might be a reason.

qasim90:
@mrexox I can print PATH contents from `sh` with `echo $PATH` and it outputs all paths correctly but still Lefthook gives this error `sh: flutter command not found` when trying to commit. What is the solution?

mrexox:
@qasim90 , could you provide more context?

- Where does `flutter` locate?
- What is your `lefthook.yml`?
- What is your `lefthook` version?

Maybe you have a public project, so I could reproduce the issue?

mrexox:
[...]

Issue #249, opened 2021-12-13
Lefthook doesn't seem to search the path


jetole:
I am trying to use ansible-lint for linting which is installed in `.venv/bin`. `.venv/bin` is part of the path. 
```
$ which lefthook 
/home/jtole/go/bin/lefthook

$ lefthook run pre-commit
Lefthook v0.7.7
RUNNING HOOKS GROUP: pre-commit

  EXECUTE > ansible-lint
 sh: 1: ansible_lint: not found


SUMMARY: (done in 0.01 seconds)
🥊  ansible-lint

$ ansible-lint --version          
ansible-lint 5.3.0 using ansible 2.12.0

$ which ansible-lint
[[ PROJECT_PATH ]]/.venv/bin/ansible-lint
```

Envek:
> ```
>  EXECUTE > ansible-lint
> sh: 1: ansible_lint: not found
> ```

Looks like you have a typo: underscore instead of dash: `which ansible-lint` but `ansible_lint: not found`.

But it can be related to using GUI tools, that have problems with `$PATH`. See https://github.com/evilmartians/lefthook/issues/178 for example (especially https://github.com/evilmartians/lefthook/issues/178#issuecomment-833305335)

mrexox:
This issue be solved via [`rc`](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#rc) option. Probably a tweak to PATH should be applied.

PR #392, opened 2022-12-04
feature: Allow providing rc file

mrexox:
Closes https://github.com/evilmartians/lefthook/pull/185 https://github.com/evilmartians/lefthook/issues/178 https://github.com/evilmartians/lefthook/issues/99 https://github.com/evilmartians/lefthook/issues/253 https://github.com/evilmartians/lefthook/issues/249

**:zap: Summary**

- [x] Add `rc` global option which allows to set the ENVs file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- feature: Allow providing rc file ([PR #392](https://github.com/evilmartians/lefthook/pull/392) by @mrexox)
+
 ## 1.2.3 (2022-11-30)
 
 - feature: Expand env variables ([PR #391](https://github.com/evilmartians/lefthook/pull/391) by @mrexox)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -7,6 +7,7 @@
   - [`skip_output`](#skip_output)
   - [`source_dir`](#source_dir)
   - [`source_dir_local`](#source_dir_local)
+  - [`rc`](#rc)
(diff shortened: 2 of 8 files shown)


skryukov (approved):

PikachuEXE:
One issue 
The generated code in hook looks like this

```sh
set -a
[ -f '~/.lefthookrc' ] && source '~/.lefthookrc'
```
However `~` would NOT be resolved to `$HOME`

I tested with more variants:
```sh
# works
[ -f "$HOME/.lefthookrc" ] && echo "$HOME/.lefthookrc exists"
# nope
[ -f '$HOME/.lefthookrc' ] && echo '$HOME/.lefthookrc exists'
[ -f '~/.lefthookrc' ] && echo '~/.lefthookrc exists'
# Original generated one
[ -f '~/.lefthookrc' ] && source '~/.lefthookrc'
```

My current workaround is to put absolute path in `rc`:
```yml
# See following doc for details
# https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#rc

# You can choose whatever name you want.
# You can share it between projects where you use lefthook.
# Make sure the path is absolute.
rc: /Users/username/.lefthookrc
```

If this would be fixed then the doc can remain unchanged
If not the doc should be updated to not use `~/.lefthookrc`

mrexox:
Hey @PikachuEXE! You are right, I missed that thing. Although I still feel like having able to set `~` or `$HOME` would be better, so I'll just remove quoting in a template.

PR #414, opened 2023-01-07
fix: Make info dir when it is absent

### LINKED ISSUE #413: `lefthook install` fails when `.git/info` dir is missing
### :wrench: Summary

`lefthook install` fails when `.git/info` dir is missing. Workaround is that you `mkdir .git/info` before running it.

I reckon it a bug as the info dir is not provisioned by default when you just `git init` in git version 2.39.0. Although I'm not really sure if this is specific to my environment or if there's config by which the info dir is automatically made available, I'd like the fresh installation to run agnostic about it.

### Lefthook version

1.2.6 7fa22dd309e2501fd44fb8e0ccbd130a206617c8

### Steps to reproduce

In any directory run `git init` and `lefthook install`.

### Expected results

```
% lefthook install
SYNCING
SERVED HOOKS: pre-commit, pre-push
```

### Actual results

```
SYNCING
Error: open .git/info/lefthook.checksum: no such file or directory
```

### Possible Solution

I have a working fix and am going to open a PR followingly.

sato11:
Closes #413

#### :zap: Summary

The presence of `.git/info` directory has been taken for granted, but it can be actually nonexistent. This results in an unhandled enoent. I've changed it so the presence of it is made sure when initializing Repository. 

The point I believe is that `.git/info` could be absent. On the other hand I'm not sure if we can take for granted that `.git/hooks` always exists, or the same mishap can occur. At least in my environment the hooks dir is provisioned automatically (where the info is not) but I haven't been able to reason about it documentationally.

This does fix my problems but since I'm not really sure what the scope of fix should be, any feedback is welcome. Thanks in advance!

- [x] Check locally
- [ ] Add tests

### DIFF
--- a/internal/git/repository.go
+++ b/internal/git/repository.go
@@ -18,6 +18,7 @@ const (
 	cmdStagedFiles = "git diff --name-only --cached"
 	cmdAllFiles    = "git ls-files --cached"
 	cmdPushFiles   = "git diff --name-only HEAD @{push} || git diff --name-only HEAD master"
+	infoDirMode    = 0o775
 )
 
 // Repository represents a git repository.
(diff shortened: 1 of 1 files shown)


mrexox:
Hey! Thank you for the PR! I think this is really helpful! I never thought that Git can omit creating info directory. I'll release the fix soon today or tomorrow

PR #458, opened 2023-03-23
fix: make hook template compatible with shells without source command

mdesantis:
#### :zap: Summary

On Linux Ubuntu 22.04, `/bin/sh` doesn't support `source` command, so if you run this template you get an error like the following one:

```
> git commit -m wip
.git/hooks/pre-commit: 8: source: not found
```

Replacing `source` with `.` fixed it. Also, [`dot` seems to be the POSIX standard](https://stackoverflow.com/q/11588583).

- [ ] Check locally
- [ ] Add tests


mdesantis:
Just tested, works great ;)

Issue #315, opened 2022-08-10
Accessing git hook arguments from `files` command


BasixKOR:
I've tried to use `post-checkout` hook and I tried to diff the changed files using `git diff --name-only {1} {0}` but I noticed this doesn't work and `{files}` just printed `files`.

BasixKOR:
Closing as it is likely a duplicate of #269.

mrexox:
Hey, @BasixKOR ! Could you provide your lefthook.yml content? #269 fixes the quoting issue on Windows, but does nothing with git arguments like {1} and {0}. 

Also {0}, {1}, and other git hook args substitutions don't happen on `files` parameter, but if you provide a usecase I can add this feature!

mrexox:
Also when you use {0} - it prints all git arguments for the hook. Use {1} for first argument and {2} for the second. See https://github.com/evilmartians/lefthook/blob/master/docs/full_guide.md#git-hook-argument-shorthands-in-commands

fabn:
> Also {0}, {1}, and other git hook args substitutions don't happen on `files` parameter, but if you provide a usecase I can add this feature!

Hi @mrexox I was looking for the same thing, I want to perform `bundle install` in a ruby project whenever a Gemfile is changed. So my idea was something about 

```yaml
post-checkout:
  commands:
    01-bundle-install:
      files: git diff --name-only {1} {2} | grep -i Gemfile
      run: bundle install
```

But is not working because `{1}` and `{2}` are not forwarded to `files`. Is there any workaround apart from writing a custom script in `run` that do that?

PR #461, opened 2023-03-31
feat: pass numeric placeholders to files command

fabn:
It address the use case outlined here

https://github.com/evilmartians/lefthook/issues/315#issuecomment-1491670663

```yaml
post-checkout:
  commands:
    01-bundle-install:
      files: git diff --name-only {1} {2}
      glob: '*Gemfile'
      run: bundle install
```

- [x] Check locally
- [ ] Add tests

I wasn't able to test it because the line I changed is not covered by current tests


mrexox:
Thank you! :rocket:

PR #462, opened 2023-03-31
feat: allow extra hooks in local config

fabn:
Without this patch is not possible to define extra hooks in local config, e.g.

```yaml
# lefthook.yml
tests:
  commands:
    tests:
      run: go test ./...
```

and 

```yml
# lefthook-local.yml
lints:
  scripts:
    "linter.sh":
      runner: bash
```

Won't work, it wont be possible to run `lefthook run lints`

#### :zap: Summary

Iterate against both global and local top level keys when adding custom hooks.

- [x] Check locally
- [x] Add tests


mrexox:
Thank you for bringing this feature! I will release a new version this week.

PR #468, opened 2023-04-11
feat: don't show when commands are skipped because of no matched files

technicalpickles:
Closes https://github.com/evilmartians/lefthook/discussions/467

#### :zap: Summary

If you have lots of commands, and make small commits, you'll often see a lot of skipped commands:

```
sorbet: (SKIP. NO FILES FOR INSPECTION)
stylelint: (SKIP. NO FILES FOR INSPECTION)
codeowners: (SKIP. NO FILES FOR INSPECTION)
authorization: (SKIP. NO FILES FOR INSPECTION)
svgo: (SKIP. NO FILES FOR INSPECTION)
syncpack: (SKIP. NO FILES FOR INSPECTION)
eslint: (SKIP. NO FILES FOR INSPECTION)
packwerk: (SKIP. NO FILES FOR INSPECTION)
rubocop: (SKIP. NO FILES FOR INSPECTION)
```

I wanted a way to skip this. I'm migrating from a home grown hook runner, and it had pretty concise output that I'm trying to recreate. I was able to do that with most of the other options, but still had this output.

I apologize in advance, this is my first go contribution to anything 🤪 

- [x] Check locally
- [ ] Add tests

### DIFF
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -351,9 +351,9 @@ func (r *Runner) runCommands() {
 func (r *Runner) runCommand(name string, command *config.Command) {
 	args, err := r.prepareCommand(name, command)
 	if err != nil {
-    if (!(r.SkipSettings.SkipSkip() && err.Error() == "no files for inspection" || err.Error() == "no matching staged_files")) {
-      logSkip(name, err.Error())
-    }
+		if !(r.SkipSettings.SkipSkip() && err.Error() == "no files for inspection" || err.Error() == "no matching staged_files") {
+			logSkip(name, err.Error())
+		}
 		return
 	}
 
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -63,7 +63,7 @@ func (r *Runner) RunAll(sourceDirs []string) {
 	}
 
 	if r.Hook.Skip != nil && r.Hook.DoSkip(r.Repo.State()) {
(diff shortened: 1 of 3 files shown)


internal/lefthook/runner/runner.go
```diff
@@ -351,7 +351,9 @@ func (r *Runner) runCommands() {
 func (r *Runner) runCommand(name string, command *config.Command) {
 	args, err := r.prepareCommand(name, command)
 	if err != nil {
-		logSkip(name, err.Error())
+		if !(r.SkipSettings.SkipSkip() && err.Error() == "no files for inspection" || err.Error() == "no matching staged_files") {
+			logSkip(name, err.Error())
+		}
```
mrexox, line 356: I think it's better put this code in `logSkip` function and make it a `Runner`'s method in order to be able to check SkipSettings. There are plenty places that log about skipping :)

mrexox (changes requested):
Thank you for contribution! This totally makes sense (although `skip skip` sounds funny). I have a small suggestion to refactor the code and put a skip settings check into a `logSkip` function. 

I'd also suggest to name the skip_output setting `skips`, so it sounds a bit more natural (like, skip the skips) since we probably want to skip all kinds of skips logging that we have in lefthook :slightly_smiling_face: 

If you don't have time for these edits I can do it for you, but I'll wait for your response :raised_hands:

mrexox (approved):
Thank you! Now everything is perfect!

PR #488, opened 2023-05-22
feat: add scoop package

mrexox:
#### TODO


sitiom:
See https://github.com/evilmartians/lefthook/issues/136#issuecomment-1646524611. Lefthook is now available in the main bucket.

PR #489, opened 2023-05-22
Allow using json configs

mrexox:
**:wrench: Summary**


jbergstroem:
Would love to see a schema to validate it as well

Issue #510, opened 2023-06-26
lefthook is artificially slow when node_modules is not in the Git root


isaaclyman:
### :wrench: Summary

On my team we have a project structured like so:

- project/
  - .git/
  - front_end/
    - package.json
    - node_modules/
  - back_end/

So our package.json and node_modules are located in a subfolder of the project root. But our pre-commit hook, handled by lefthook, was taking 45+ seconds to run on every commit, which was strange because if I ran `npx lefthook run pre-commit` it took less than 5 seconds.

I finally looked into it, added some `echo` statements to the pre-commit script, and discovered the following:

- The script looks for an installed version of lefthook in $dir/node_modules/lefthook, where $dir is the top-level directory of the Git working tree.
- If not found, eventually the script falls back to `npx @evilmartians/lefthook`.
- NPX first looks in the current directory for node_modules binaries and packages.
- It then looks in PATH.
- If not found, NPX downloads the named package from the NPM registry.
- The download and install take far more time than the hooks themselves.

There's no message to indicate that lefthook is falling back to installing itself from the NPM registry, so it took me a while to figure out what has happening.

### Lefthook version

1.4.3

### Steps to reproduce

- Create a lefthook.yml at the root of a Git repository with a pre-commit hook
- In a subfolder, run `npm install --save-dev lefthook`
- In the project root, run `npx lefthook install`
- Run `git commit`

### Expected results

I would see some message indicating that my lefthook installation was not found.

### Actual results

The version of lefthook installed in subfolder/node_modules is not found, and @evilmartians/lefthook is downloaded from the NPM registry instead.

### Possible Solutions

- A config option in lefthook.yml could be provided to indicate where the node_modules folder is located.
- The pre-commit script could search the project recursively up to a certain depth to find node_modules.
- If all previous attempts to find lefthook fail, a warning message could indicate that `npx` may download it from NPM, introducing a large delay into the hook.

montchr:
I haven't investigated the cause, but I've also been experiencing extremely long delays, oftentimes around 30 seconds per commit. I have not seen this issue when running git commands in an interactive shell, but it does occur when using Magit in Emacs. I don't know when this started exactly, but I would guess sometime in the past month or two – I had been using lefthook without issue before then.

bard:
[...]

PR #516, opened 2023-07-11
Improve hook template

hyperupcall:
Related to #510

#### :zap: Summary

This improves `hook.tmpl` by removing extra work and fixing subtle bugs in passing arguments with whitespace.

- The first commit removes a useless echo (more info can be found [here](https://www.shellcheck.net/wiki/SC2116))
- The second commit removes all "eval" builtins as they are not necessary. `$@` has been quoted in all places as should be to properly preserve whitespace.
- The third commit partially solves #510. I think it is OK to assume that `@evilmartians/lefthook` will be executed if `npx` exists. Cases where it may not may include when a directory has incorrect permissions or the registry cannot be resolved - cases in which the user has a broken configuration already. It is more frustrating when it is artificially slow, especially since `npx` kind of slow by itself.

- [x] Check locally
- [ ] Add tests

### DIFF
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -43,7 +43,7 @@ call_lefthook()
   elif pnpm lefthook -h >/dev/null 2>&1
   then
     pnpm lefthook "$@"
-  elif npx @evilmartians/lefthook -h >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


mrexox (approved):
Thank you! I will release this PR later today. This change totally makes sense!

PR #519, opened 2023-07-14
fix: Do not print extraneous newlines when executionInfo output is hidden

### LINKED ISSUE #494: Compress the blank lines printed
### :wrench: Summary

There seemed to be exceeded blank lines been printed after the hook execution

### Lefthook version

1.4.1 a422566554a1b6c1295125715ddec86d6c0de423

### Steps to reproduce

1. Clone the develop branch of this repo: https://github.com/hustcer/setup-nu/tree/develop
2. `pnpm i ` to install dependencies
3. Modify some files, say: `src/index.ts` and`src/setup.ts` 
4. Commit the changes and you will got the following output:
```console
$> git commit -am 'try format'


Formatted 2 file(s) in 12ms




[develop 80f381f] try format
 3 files changed, 2 insertions(+), 3 deletions(-)
```
There is some unnecessary blank lines

### Expected results

```console
$> git commit -am 'try format'

Formatted 2 file(s) in 12ms

[develop 80f381f] try format
 3 files changed, 2 insertions(+), 3 deletions(-)
```

### Actual results

See above

### Logs / Screenshots

```bash
LEFTHOOK_VERBOSE=true git ...
```

hyperupcall:
Closes #494

#### :zap: Summary

Prevents printing blank lines when `exececutionInfo` is not printed.

Previously, two newlines would be printed for each command (Effectively `fmt.Println("\n")`). This fixes that.

- [x] Check locally
- [ ] Add tests

### DIFF
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -485,12 +485,12 @@ func (r *Runner) logExecute(name string, err error, out io.Reader) {
 		execLog = fmt.Sprint(log.Cyan("\n  EXECUTE > "), log.Bold(name))
 	}
 
+	log.Info(execLog)
+
 	if err == nil && r.SkipSettings.SkipExecutionOutput() {
-		log.Info(execLog)
 		return
 	}
 
-	log.Info(execLog)
 	if out != nil {
 		log.Info(out)
 	}
--- a/internal/lefthook/runner/runner.go
+++ b/internal/lefthook/runner/runner.go
@@ -478,14 +478,16 @@ func (r *Runner) logExecute(name string, err error, out io.Reader) {
 	var execLog string
 	switch {
 	case r.SkipSettings.SkipExecutionInfo():
-		execLog = "\n"
+		execLog = ""
 	case err != nil:
 		execLog = fmt.Sprint(log.Red("\n  EXECUTE > "), log.Bold(name))
 	default:
 		execLog = fmt.Sprint(log.Cyan("\n  EXECUTE > "), log.Bold(name))
 	}
 
-	log.Info(execLog)
+	if execLog != "" {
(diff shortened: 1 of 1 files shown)


internal/lefthook/runner/runner.go
```diff
@@ -478,19 +478,21 @@ func (r *Runner) logExecute(name string, err error, out io.Reader) {
...
 	default:
 		execLog = fmt.Sprint(log.Cyan("\n  EXECUTE > "), log.Bold(name))
 	}
 
-	if err == nil && r.SkipSettings.SkipExecutionOutput() {
+	if execLog != "" {
 		log.Info(execLog)
+	}
+
+	if err == nil && r.SkipSettings.SkipExecutionOutput() {
```
mrexox, line 492: These two logs should be switched since we must skip logs execution here.

mrexox (commented):
Overall looks good, just a small nitpick.

internal/lefthook/runner/runner.go
hyperupcall: Done :+1: 

So I guess this means that the previous [`log.Info(execLog)`](https://github.com/evilmartians/lefthook/blob/9f47c64faff4c4f417c3dcf578e8bf592c36ae9b/internal/lefthook/runner/runner.go#L489) was not intended to be there - Because it seems that now the behavior just changed so that `execLog` is never printed when the condition `err == nil && r.SkipSettings.SkipExecutionOutput()` is met.

mrexox: Oh god, I'm sorry. `execLog` shouldn't be printed if it is empty. That's right. And if we have SkipExecutionOutput() then we don't print command output. So, your previous solution was correct. Let me fix that.

mrexox (approved):

PR #520, opened 2023-07-14
Support `.lefthook.yml` and `.lefthook-local.yml`

### LINKED ISSUE #319: support new config file format
It will be really cool to add `.lefthook.yml` (with dot in the beginning) config file format support. Just to align with other libs:
- [lint-staged](https://github.com/okonet/lint-staged#configuration)
- [syncpack](https://github.com/JamieMason/syncpack#-configuration-file)
- [eslint](https://eslint.org/docs/latest/user-guide/configuring/configuration-files)

hyperupcall:
Closes #319

#### :zap: Summary

This supports the new `.lefthook*` formats:

- As noted in the issue, the dot-prefixed and non-dot-prefixed files do not have any overriding behaviors. The "global" and local files are also resolved separately.
- The default was also changed to automatically generate a `.lefthook.yml` if one doesn't already exist.
- Local config uses `-local` (`.local` was suggested, but this is easier and in line with the scripts directory)
- If both the dot-prefixed and non-dot-prefixed files exist for either "global" or local, no errors are thrown (this was suggested, I did not implement it).

I was wondering some other things:

- In the [uninstall command](https://github.com/evilmartians/lefthook/blob/c134f31b80547ecc14f26581c3c3f7f5c151087d/internal/lefthook/uninstall.go#L36), I modified things to remove these new config files. Since lefthook supports JSON and TOML files, shouldn't this remove those files as well?

- [x] Check locally
- [x] Add tests

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -34,13 +34,8 @@ func (err NotFoundError) Error() string {
 
 // Loads configs from the given directory with extensions.
 func Load(fs afero.Fs, repo *git.Repository) (*Config, error) {
-	global, err := read(fs, repo.RootPath, "lefthook")
+	global, err := readOne(fs, repo.RootPath, []string{".lefthook", "lefthook"})
 	if err != nil {
-		var notFoundErr viper.ConfigFileNotFoundError
-		if ok := errors.As(err, &notFoundErr); ok {
-			return nil, NotFoundError{err.Error()}
-		}
-
 		return nil, err
 	}
 
@@ -81,9 +76,27 @@ func read(fs afero.Fs, path string, name string) (*viper.Viper, error) {
 	return v, nil
 }
 
+func readOne(fs afero.Fs, path string, names []string) (*viper.Viper, error) {
+	for _, name := range names {
+		v, err := read(fs, path, name)
+		if err != nil {
+			var notFoundErr viper.ConfigFileNotFoundError
+			if ok := errors.As(err, &notFoundErr); ok {
+				continue
+			} else {
+				return nil, err
+			}
+		}
+
+		return v, nil
+	}
+
+	return nil, NotFoundError{fmt.Sprintf("No config files with names %q could not be found in \"%s\"", names, path)}
+}
+
 // mergeAll merges remotes and extends from .lefthook and .lefthook-local.
 func mergeAll(fs afero.Fs, repo *git.Repository) (*viper.Viper, error) {
-	extends, err := read(fs, repo.RootPath, "lefthook")
+	extends, err := readOne(fs, repo.RootPath, []string{".lefthook", "lefthook"})
 	if err != nil {
 		return nil, err
 	}
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -35,7 +35,7 @@ func (err NotFoundError) Error() string {
 
 // Loads configs from the given directory with extensions.
 func Load(fs afero.Fs, repo *git.Repository) (*Config, error) {
-	global, err := readOne(fs, repo.RootPath, []string{".lefthook", "lefthook"})
+	global, err := readOne(fs, repo.RootPath, []string{"lefthook", ".lefthook"})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


hyperupcall:
codecov seemed to decrease but new missed lines seem to be similar to existing ones?

internal/config/load.go
```diff
@@ -81,9 +77,28 @@ func read(fs afero.Fs, path string, name string) (*viper.Viper, error) {
...
+	}
+
+	return nil, NotFoundError{fmt.Sprintf("No config files with names %q could not be found in \"%s\"", names, path)}
+}
+
+// mergeAll merges (.lefthook or lefthook) and (extended config) and (remote)
+// and (.lefthook-local or .lefthook-local) configs.
 func mergeAll(fs afero.Fs, repo *git.Repository) (*viper.Viper, error) {
-	extends, err := read(fs, repo.RootPath, "lefthook")
+	extends, err := readOne(fs, repo.RootPath, []string{".lefthook", "lefthook"})
```
mrexox, line 101: I'd like to make dot variant (`.lefthook.yml`) just an optional supported one. So, let's change this line to check `lefthook.yml` first.
```suggestion
	extends, err := readOne(fs, repo.RootPath, []string{"lefthook", ".lefthook"})
```

internal/config/load.go
```diff
@@ -96,7 +111,7 @@ func mergeAll(fs afero.Fs, repo *git.Repository) (*viper.Viper, error) {
 		return nil, err
 	}
 
-	if err := merge("lefthook-local", "", extends); err == nil {
+	if err := mergeOne([]string{".lefthook-local", "lefthook-local"}, "", extends); err == nil {
```
mrexox, line 114: Let's do the same here. 

```suggestion
	if err := mergeOne([]string{"lefthook-local", ".lefthook-local"}, "", extends); err == nil {
```

internal/lefthook/install.go
```diff
@@ -99,7 +99,7 @@ func (l *Lefthook) configExists(path string) bool {
 }
 
 func (l *Lefthook) createConfig(path string) error {
-	file := filepath.Join(path, config.DefaultConfigName)
+	file := filepath.Join(path, config.DefaultDotConfigName)
```
mrexox, line 102: Let's keep this behavior unchanged? We'll change this in later major release. I think it would be great to just support `.lefthook.yml` and `.lefthook-local.yml` by now, fix possible issue, decide on what to do when both variants exist and so on. So, this transition will be smooth. WDYT?

mrexox:
> The default was also changed to automatically generate a .lefthook.yml if one doesn't already exist.

Let's keep this behavior unchanged? Then we can release this PR with just a minor/patch version and polish it before making `.lefthook.yml` the default.

> In the [uninstall command](https://github.com/evilmartians/lefthook/blob/c134f31b80547ecc14f26581c3c3f7f5c151087d/internal/lefthook/uninstall.go#L36), I modified things to remove these new config files. Since lefthook supports JSON and TOML files, shouldn't this remove those files as well?

You are right! I will fix it myself, you don't have to fix this in this PR, thank you for noticing!

internal/lefthook/install.go
hyperupcall: That sounds good. The docs don't mention of the dot-prefixed variants so there is no change of confusion if one is generated or has priority over the non-dot-prefixed variants.

internal/config/load_test.go
```diff
@@ -614,12 +614,16 @@ pre-push:
 		}
 
 		t.Run(fmt.Sprintf("%d: %s", i, tt.name), func(t *testing.T) {
-			if err := fs.WriteFile(filepath.Join(root, "lefthook.yml"), []byte(tt.global), 0o644); err != nil {
-				t.Errorf("unexpected error: %s", err)
+			if tt.global != "" {
```
hyperupcall, line 617: Only write to the config file if we explicitly specify its content. That way, if `.lefthook.yml` is written to, an empty `lefthook.yml` doesn't take precedence

mrexox (approved):

Issue #76, opened 2019-09-26
Skip symlinks?


nikolay:
We have tons of symlinks in a project. Is it possible to add a filter to skip symlinks in `{staged_files}`, `{all_files}`, etc., so that the same file won't be linted twice when the symlinks gets resolved?

Arkweid:
Yep, for sure. I think I will not have a problem with this.

sanmai-NL:
Is there a development on this issue? Prettier exits with an error exit status if a symlink is between, e.g. `{all_files}`.

sanmai-NL:
@mrexox Yes, see https://github.com/evilmartians/lefthook/issues/76 for a well-argued use case as solution direction.

I don't know if a boolean would be optimal, since a Git file mode filter would be more flexible.

mrexox:
Please, check out **1.6.10** release. There's a [`file_types`](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#file_types) option added. You can use it to filter the symlinks:

```yml
# lefthook.yml

pre-commit:
  commands:
    lint:
      run: npm run lint {all_files}
      file_types: not symlink
```

PR #538, opened 2023-08-15
feat: add skip_symlinks option to commands

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/76

**:zap: Summary**


mrexox:
This feature will be a part of the https://github.com/evilmartians/lefthook/pull/698 PR with filters on file types.

PR #543, opened 2023-08-29
fix: allow empty staged files diffs

mrexox:
**:wrench: Summary**


scop:
This seems to have introduced the problem in #584.

PR #552, opened 2023-09-22
fix: Improve script performance and reliability

### LINKED ISSUE #249: Lefthook doesn't seem to search the path

### LINKED ISSUE #510: lefthook is artificially slow when node_modules is not in the Git root

hyperupcall:
~~Previously, only lefthook stuff under the current `$PWD/node_modules` was checked. Now, the current directory up until the Git root is checked (see `try_exec` function.~~

I did chang the formatting since it was really difficult to parse the shell script. If you want to change it back, let me know.

#### :zap: Summary

Optimizations:

- ~~Don't waste time running a subshell with `git rev-parse --show-toplevel` when it is not needed~~
- Does not attempt to run `lefthook` to determine if `lefthook` can be run (use `command -v` instead)'
  - slight performance improvement 
- Use `exec` to run commands
  - Makes process signals more reliable (so that they are not ignored by the shell)

~~Note that this slightly changes the resolution algorithm. Before, it would look for `./node_modules/lefthook/bin/index.js`, then `./node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}`, etc. Now, it looks for `./node_modules/lefthook/bin/index.js`, then `../node_modules/lefthook/bin/index.js`, etc., all the way up to the git root. Once it reaches git root, _then_ it goes back to the original `$PWD` and looks for `./node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}`, and climbs up the directory tree for that path.~~

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -9,45 +9,33 @@ fi
 [ -f {{.Rc}} ] && . {{.Rc}}
 {{- end}}
 
-call_lefthook()
-{
-  dir="$(git rev-parse --show-toplevel)"
+call_lefthook() {
   osArch=$(uname | tr '[:upper:]' '[:lower:]')
   cpuArch=$(uname -m | sed 's/aarch64/arm64/')
 
-  if lefthook{{.Extension}} -h >/dev/null 2>&1
-  then
-    lefthook{{.Extension}} "$@"
+  if command -v lefthook{{.Extension}} >/dev/null 2>&1; then
+    exec lefthook{{.Extension}} "$@"
   {{if .Extension -}}
   {{/* Check if lefthook.bat exists. Ruby bundler creates such a wrapper */ -}}
-  elif lefthook.bat -h >/dev/null 2>&1
-  then
-    lefthook.bat "$@"
+  elif command -v lefthook.bat >/dev/null 2>&1; then
+    exec lefthook.bat "$@"
   {{end -}}
-  elif test -f "$dir/node_modules/lefthook/bin/index.js"
-  then
-    "$dir/node_modules/lefthook/bin/index.js" "$@"
-  elif test -f "$dir/node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}"
-  then
-    "$dir/node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}" "$@"
-  elif test -f "$dir/node_modules/@evilmartians/lefthook-installer/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}"
-  then
-    "$dir/node_modules/@evilmartians/lefthook-installer/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}" "$@"
-  elif bundle exec lefthook -h >/dev/null 2>&1
-  then
-    bundle exec lefthook "$@"
-  elif yarn lefthook -h >/dev/null 2>&1
-  then
-    yarn lefthook "$@"
-  elif pnpm lefthook -h >/dev/null 2>&1
-  then
-    pnpm lefthook "$@"
-  elif command -v npx >/dev/null 2>&1
-  then
-    npx @evilmartians/lefthook "$@"
-  elif swift package plugin lefthook >/dev/null 2>&1
-  then
-    swift package --disable-sandbox plugin lefthook "$@"
+  fi
+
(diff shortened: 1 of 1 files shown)


mrexox:
Thank you for submitting a PR! I will take a look and come back in a few days.

mrexox:
Are you sure this will fix https://github.com/evilmartians/lefthook/issues/510? In the issue they have `node_modules` in a subdir, not the parent dir 🤔

hyperupcall:
Oh oops, you're right, it doesn't fix that issue. I was having issues where the existing lefthook binary couldn't be found because I was in a subdirectory of the root git directory - and I did not read that issue carefully enough and thought it was the same problem.

mrexox:
@hyperupcall , I've just reviewed the PR one more time. I can't think of a case when we need to check parent directories to find the lefthook executable in node_modules. What do you think about removing this part? Other optimizations look good to me.

hyperupcall:
@mrexox Sounds good - I have removed the recursive parent directory search. I have kept the other optimizations

internal/templates/hook.tmpl
```diff
@@ -9,45 +9,33 @@ fi
...
-  then
-    npx @evilmartians/lefthook "$@"
-  elif swift package plugin lefthook >/dev/null 2>&1
-  then
-    swift package --disable-sandbox plugin lefthook "$@"
+  fi
+
+  try_exec "./node_modules/lefthook/bin/index.js" "$@"
+  try_exec "./node_modules/@evilmartians/lefthook/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}" "$@"
+  try_exec "./node_modules/@evilmartians/lefthook-installer/bin/lefthook_${osArch}_${cpuArch}/lefthook{{.Extension}}" "$@"
```
mrexox, line 27: There was $dir variable assigned before. Since we don't check upper directories, please, add this variable

hyperupcall:
Because the original changes of this PR did not fully fix the mentioned issue, and the current diff no longer reflecting the original issue, and the length of time that has passed since creating this issue, I'm closing this PR. If needed, someone else can pick up the torch :fire: - a fresh PR would probably be easier to review too

PR #553, opened 2023-09-22
fix: Do not enable export when sourcing rc file

hyperupcall:
Doesn't close any issue.

#### :zap: Summary

There's currently a discrepancy with the documentation on [rc](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#rc) files and the actual behavior.

The documentation states:

> Provide a tweak to access npm executable the same way you do it in your ~/rc

However, `set -a` is not set by default (or at all in most people's configurations) in `~/.profile`, `~/.bashrc`, etc. If that were the case, then way too many environment variables would be added to the environment table of all child processes.

The correct behavior is already hinted at with the examples:

```sh
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```

The `export` syntax is used, which implies that `set -a` is not enabled. Furthermore, when sourcing `"$NVM_DIR/nvm.sh"`, we wouldn't want all the regular variables (not environment variables) it defines to be loaded in as environment variables - again this behavior is different from standard rc files and is unintuitive. Other hook managers like [Husky](https://github.com/typicode/husky/blob/main/husky.sh) act like this.

This is a breaking change, but is in line with user expectations, the current documentation, and other popular Git hook managers.

This also adds documentation to make it clear how to properly use `rc` field when path contains spaces.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -243,7 +243,7 @@ pre-commit:
       run: npm run eslint {staged_files}
 ```
 
-Provide a tweak to access `npm` executable the same way you do it in your ~/<shell>rc
+Provide a tweak to access `npm` executable the same way you do it in your ~/<shell>rc.
 
 ```yml
 # lefthook-local.yml
@@ -254,16 +254,25 @@ Provide a tweak to access `npm` executable the same way you do it in your ~/<she
 rc: ~/.lefthookrc
 ```
 
+Or
+
+```yml
+# lefthook-local.yml
+
+# If the path contains spaces, you need to quote it.
+rc: '"${XDG_CONFIG_HOME:-$HOME/.config}/lefthookrc"'
+```
+
+In the rc file, export any new environment variables or modify existing ones.
+
 ```bash
 # ~/.lefthookrc
 
 # An nvm way
-
 export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
 # Or maybe just
-
 PATH=$PATH:$HOME/.nvm/versions/node/v15.14.0/bin
 ```
 
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -5,8 +5,7 @@ if [ "$LEFTHOOK" = "0" ]; then
 fi
 
 {{- if .Rc}}
-{{/* Load rc file and export all ENV vars defined in it */}}
-set -a
+{{/* Load rc file, which may export ENV variables */}}
 [ -f {{.Rc}} ] && . {{.Rc}}
 {{- end}}
 


docs/configuration.md
```diff
@@ -254,16 +254,25 @@ Provide a tweak to access `npm` executable the same way you do it in your ~/<she
...
 # ~/.lefthookrc
 
 # An nvm way
-
 export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
 # Or maybe just
-
 PATH=$PATH:$HOME/.nvm/versions/node/v15.14.0/bin
```
mrexox, line 276: We probably need to change this line to be `export PATH=$PATH:$HOME/.nvm/versions/node/v15.14.0/bin`

hyperupcall: @mrexox `PATH` is already an environment variable (that's set by `login`, etc.):

```sh
PATH=bar /usr/bin/sh -c 'export -p | /usr/bin/grep "export PATH="'
export PATH='bar'

PATH=bar /usr/bin/bash -c 'declare -p PATH'     
declare -x PATH="bar"
```

hyperupcall:
> Do you think set -a will add too much ENVs? What is the price and is it really significant?

About the actual number, it could set a few ENV varaibles, or a few hundred - but the former is more likely and the latter is likely quite rare. Shell utilities like [nvm](https://github.com/nvm-sh/nvm/blob/master/nvm.sh) and [autoenv](https://github.com/hyperupcall/autoenv/blob/master/activate.sh) generally set many functions, and relatively few variables (that would become environment varaibles due to `set -a`). This is because most of these tools use non-standard builtins like `local`, which confines the exported environment varaible to the function-scope (most strictly POSIX-compatible shells implement `local` as a convenience). There are exceptions (ex. [pfetch](https://github.com/dylanaraps/pfetch/blob/master/pfetch)), but they are quite rare. In my opinion, 1 variable that is unecessarily exported is too many variables.

The price is only correctness, and preformance is negligible. At first, it only seems like an inconvenience when variables like `usage` or `_pwd` are exported into the environment. But eventually, and given enough time, people will use variables like `NAME`, `EMAIL`, `EDITOR`, etc. (i still see people use non-exported, uppercase varaibles in the wild), and that will leak into all scripts `lefthook` runs, and people will wonder commands ran under `lefthook` behave differently.

hyperupcall:
@mrexox Were you planning on merging this change? - `testscript` seems to be behaving now...
If not, then I can close this issue

mrexox:
> @mrexox Were you planning on merging this change? - `testscript` seems to be behaving now... If not, then I can close this issue

Yes, it looks good. I'll merge it.

Issue #560, opened 2023-10-05
Separator sometimes shown when there isn't something to separate


technicalpickles:
### :wrench: Summary

I have quite a few `skip_output` options. With the specific combination I use, I've found that there's an extra separator displayed, even though there isn't any prior output to separate

### Lefthook version

1.5.0

### Steps to reproduce

Here's my `skip_output` configuration:

```yaml
skip_output:
  - meta           # Skips lefthook version printing
  # - summary      # Skips summary block (successful and failed steps) printing
  - empty_summary  # Skips summary heading when there are no steps to run
  # - success      # Skips successful steps printing
  # - failure      # Skips failed steps printing
  - execution      # Skips printing any execution logs (but prints if the execution failed)
  - execution_out  # Skips printing execution output (but still prints failed commands output)
  - execution_info # Skips printing `EXECUTE > ...` logging
  - skips          # Skips "skip" printing (i.e. no files matched)
  - meta           # Skips lefthook version printing
```

Then `lefthook run <stage>`

### Expected results

```
❯ git push
summary: (done in 9.96 seconds)
✔️  codeowners
```

### Actual results

```
❯ git push

  ────────────────────────────────────
summary: (done in 9.96 seconds)
✔️  codeowners
```

### Possible Solution

I tracked the code outputting this to:
https://github.com/evilmartians/lefthook/blob/c8e4b13eb99bdf02482e6d4bfb4f44a551320162/internal/log/log.go#L275-L289
https://github.com/evilmartians/lefthook/blob/c8e4b13eb99bdf02482e6d4bfb4f44a551320162/internal/lefthook/run.go#L184C3-L184C3

I think one way to fix would be to keep track if there was any output displayed before the summary before calling `Separate`.

### Logs / Screenshots
![CleanShot 2023-10-04 at 16 18 10@2x](https://github.com/evilmartians/lefthook/assets/159/1eab2d6a-6462-4937-a78f-8c76cbaa8b32)

mrexox:
Hey! Sorry for the long delay. What do you think if we skip summary separator when `execution` or `execution_info` and `execution_out` are set? Sounds like these are the only cases when summary separator is not needed 🤔

technicalpickles:
That makes sense to me 👍🏻 I only suggested the config option just because I couldn't trace when it was it was actually needed 😅

mrexox:
I'll release the change next week. Will probably add a few more fixes to the next release :+1:

[...]

PR #570, opened 2023-11-02
Add skip option for the summary separator

technicalpickles:
Closes https://github.com/evilmartians/lefthook/issues/560

#### :zap: Summary

Adds yet another `skip` option 😅  this time for the separator before a summary. I don't know if there is a point when there are too many skips, and what that point is though...

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/lefthook/run.go
+++ b/internal/lefthook/run.go
@@ -195,9 +195,13 @@ func printSummary(
 		return
 	}
 
-	log.Separate(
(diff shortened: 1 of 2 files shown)


technicalpickles:
Closing re https://github.com/evilmartians/lefthook/issues/560#issuecomment-1807835351

> What do you think if we skip summary separator when execution or execution_info and execution_out are set? Sounds like these are the only cases when summary separator is not needed

PR #594, opened 2023-12-14
Support `--all-including-untracked-files`

sanmai-NL:
‘All files’ currently means all files in the Git index, so this misses untracked, but unignored additions.

- [x] Check locally
- [ ] Add tests
- [x] Add documentation

### DIFF
--- a/internal/config/files.go
+++ b/internal/config/files.go
@@ -3,10 +3,11 @@ package config
 import "strings"
 
 const (
-	SubFiles       string = "{files}"
-	SubAllFiles    string = "{all_files}"
-	SubStagedFiles string = "{staged_files}"
-	PushFiles      string = "{push_files}"
+	SubFiles                      string = "{files}"
+	SubAllFiles                   string = "{all_files}"
+	SubAllFilesIncludingUntracked string = "{all_including_untracked_files}"
+	SubStagedFiles                string = "{staged_files}"
+	PushFiles                     string = "{push_files}"
 )
 
 func isRunnerFilesCompatible(runner string) bool {
--- a/internal/lefthook/run/prepare_command.go
+++ b/internal/lefthook/run/prepare_command.go
@@ -79,14 +79,17 @@ func (r *Runner) buildRun(command *config.Command) (*run, error, error) {
 		stagedFiles = func() ([]string, error) { return r.Files, nil }
 	case r.AllFiles:
 		stagedFiles = r.Repo.AllFiles
+	case r.AllFilesIncludingUntracked:
+		stagedFiles = r.Repo.AllFilesIncludingUntracked
 	default:
 		stagedFiles = r.Repo.StagedFiles
 	}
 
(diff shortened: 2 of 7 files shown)


mrexox:
Hey @sanmai-NL, could you please describe the use case? Why do you need to run lefthook hooks against untracked files?

sanmai-NL:
@mrexox As a DevOps specialist, I want data generated as part of a CI/CD pipeline to be included in checks orchestrated by Lefthook, so that I can catch suboptimalities in generation source code.

As a Developer, I want be able to run Lefthook on my entire working directory even as new content is gradually being added, without having to add and potentially remove new files to the index every time.

mrexox:
Do you use pre-commit hook for that or a custom hook name?

sanmai-NL:
A custom hook name.

sanmai-NL:
I addressed the cyclomatic complexity (only lint failure) in #595. After the review is finished for both this PR and #595, I hope it's possible to merge this as is, as the following PR #595 would be merged soon thereafter.

internal/config/files.go
```diff
@@ -3,10 +3,11 @@ package config
 import "strings"
 
 const (
-	SubFiles       string = "{files}"
-	SubAllFiles    string = "{all_files}"
-	SubStagedFiles string = "{staged_files}"
-	PushFiles      string = "{push_files}"
+	SubFiles                      string = "{files}"
+	SubAllFiles                   string = "{all_files}"
+	SubAllFilesIncludingUntracked string = "{all_including_untracked_files}"
```
mrexox, line 8: @sanmai-NL , do you really need this template? Why can't a custom `{files}` be used instead?

internal/lefthook/run/prepare_command.go
```diff
@@ -79,14 +79,17 @@ func (r *Runner) buildRun(command *config.Command) (*run, error, error) {
 		stagedFiles = func() ([]string, error) { return r.Files, nil }
 	case r.AllFiles:
 		stagedFiles = r.Repo.AllFiles
+	case r.AllFilesIncludingUntracked:
+		stagedFiles = r.Repo.AllFilesIncludingUntracked
```
mrexox, line 83: If there's no need to use `{all_including_untracked_files}` template we can move this part to `internal/lefthook/run.go` and pass those files via `Files` option. I suggested pretty similar solution [here](https://github.com/evilmartians/lefthook/pull/596#discussion_r1429726613)

mrexox:
This PR is pretty stale. Please, create an issue, and I will prepare this change if you still find it useful.

Right now I see two workarounds:

1. Add custom `files: git ls-files --cached --others --exclude-standard` option to list all files and use `{files}` template in `run` option.
2. Use `lefthook run <hook> --files-from-stdin` and provide the desired list of files as STDIN for lefthook command.

I agree that it can be useful to list all files, even untracked. So, I'll be glad to think out this feature.

PR #596, opened 2023-12-15
More robust collection of file paths

### LINKED ISSUE #564: `--files` path separator  not specified, whitespace doesn't work so it doesn't process globs
### :wrench: Summary

`--files` is supposed to take a list of file paths, but it isn't specified what the separator between these paths should be. A pretty common use case is glob patterns in a shell. Those don't work, only the first path is taken. 
### Lefthook version

1.5.2

### Steps to reproduce

```yml
pre-push:
  commands: &pre-push
    bandit:
      files: git ls-files pyproject.toml
      glob: '*.py'
      run: echo bandit --configfile pyproject.toml --quiet -- {staged_files}
```

```console
$ echo **.py
src/__version__.py src/package/__main__.py src/package/__init__.py
```

```console
$ lefthook run pre-push --commands bandit --files src/**.py --force
╭────────────────────────────────────╮
│ 🥊 lefthook v1.5.2  hook: pre-push │
╰────────────────────────────────────╯
┃  bandit ❯ 

bandit --configfile pyproject.toml --quiet -- src/__version__.py
```

### Expected results

Bandit is being run with all paths matched by the glob.

### Actual results

Bandit is being run with the first path matched by the glob.

### Possible Solution

Reuse `$IFS` is the input separator under Unix, a similar variable or whitespace under Windows, and offer a `-0` option to use the null byte as file path separator instead.

### Logs / Screenshots

```bash
LEFTHOOK_VERBOSE=true git ...
```

sanmai-NL:
Closes #564.

#### :zap: Summary

Lefthook is currently not completely ready for repositories with filenames that contain commas.

1. Add `--file` as the successor of `--files`. This option can be repeated for a list of file paths. `--files` is dangerous, e.g., one can create a file `README.md,..` and if it gets supplied to a fixer, who knows what happens.
2. Add `--files-from-stdin`. This option reads file paths from stdin, separated by a null character. This allows for file names with a comma or newline, and is interoperable with Git and Unix tools (e.g., `git ls-files -z`, `git diff --name-only -z`, `find -print0`).

@mrexox Can you test and add automated tests? I specifically advise to test interaction with existing stdin-related features.

- [x] Check locally
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/evilmartians/lefthook/internal/lefthook"
+	"github.com/evilmartians/lefthook/internal/log"
 )
 
 func newRunCmd(opts *lefthook.Options) *cobra.Command {
@@ -41,7 +42,12 @@ func newRunCmd(opts *lefthook.Options) *cobra.Command {
 		"get files from standard input, null-separated, instead of from CLI parameters --file(s)",
 	)
 
-	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "[DEPRECATED] run on specified files, comma-separated. takes precedence over --all-files")
+	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "run on specified files, comma-separated. takes precedence over --all-files")
+
+	err := runCmd.Flags().MarkDeprecated("files", "use --file flag instead")
+	if err != nil {
+		log.Warn("Unexpected error:", err)
+	}
 
 	runCmd.Flags().StringArrayVar(&runArgs.Files, "file", nil, "run on specified file (repeat for multiple files). takes precedence over --all-files")
 
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -36,6 +36,11 @@ func newRunCmd(opts *lefthook.Options) *cobra.Command {
 		"run hooks on all files",
 	)
 
+	runCmd.Flags().BoolVar(
+		&runArgs.FilesFromStdin, "files-from-stdin", false,
+		"get files from standard input, null-separated, instead of from CLI parameters --file(s)",
+	)
+
 	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "[DEPRECATED] run on specified files, comma-separated. takes precedence over --all-files")
 
 	runCmd.Flags().StringArrayVar(&runArgs.Files, "file", nil, "run on specified file (repeat for multiple files). takes precedence over --all-files")
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -39,19 +39,28 @@ func newRunCmd(opts *lefthook.Options) *cobra.Command {
 
 	runCmd.Flags().BoolVar(
 		&runArgs.FilesFromStdin, "files-from-stdin", false,
-		"get files from standard input, null-separated, instead of from CLI parameters --file(s)",
+		"get files from standard input, null- or \\n-separated",
 	)
 
-	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "run on specified files, comma-separated. takes precedence over --all-files")
+	runCmd.Flags().StringSliceVar(
+		&runArgs.Files, "files", nil,
+		"run on specified files, comma-separated",
+	)
+
+	runCmd.Flags().StringArrayVar(
+		&runArgs.Files, "file", nil,
+		"run on specified file (repeat for multiple files). takes precedence over --all-files",
+	)
+
+	runCmd.Flags().StringSliceVar(
+		&runArgs.RunOnlyCommands, "commands", nil,
+		"run only specified commands",
+	)
 
 	err := runCmd.Flags().MarkDeprecated("files", "use --file flag instead")
 	if err != nil {
 		log.Warn("Unexpected error:", err)
 	}
-
-	runCmd.Flags().StringArrayVar(&runArgs.Files, "file", nil, "run on specified file (repeat for multiple files). takes precedence over --all-files")
-
-	runCmd.Flags().StringSliceVar(&runArgs.RunOnlyCommands, "commands", nil, "run only specified commands")
 
 	return &runCmd
 }
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -36,7 +36,9 @@ func newRunCmd(opts *lefthook.Options) *cobra.Command {
 		"run hooks on all files",
 	)
 
-	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "run on specified files. takes precedence over --all-files")
+	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "[DEPRECATED] run on specified files, comma-separated. takes precedence over --all-files")
+
+	runCmd.Flags().StringArrayVar(&runArgs.Files, "file", nil, "run on specified file (repeat for multiple files). takes precedence over --all-files")
 
 	runCmd.Flags().StringSliceVar(&runArgs.RunOnlyCommands, "commands", nil, "run only specified commands")
 
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -25,10 +25,11 @@ type Command struct {
 	Exclude  string `mapstructure:"exclude"  yaml:",omitempty" json:"exclude,omitempty"  toml:"exclude,omitempty"`
 	Priority int    `mapstructure:"priority" yaml:",omitempty" json:"priority,omitempty" toml:"priority,omitempty"`
 
-	FailText    string `mapstructure:"fail_text"   yaml:"fail_text,omitempty"   json:"fail_text,omitempty"   toml:"fail_text,omitempty"`
-	Interactive bool   `mapstructure:"interactive" yaml:",omitempty"            json:"interactive,omitempty" toml:"interactive,omitempty"`
-	UseStdin    bool   `mapstructure:"use_stdin"   yaml:",omitempty"            json:"use_stdin,omitempty"   toml:"use_stdin,omitempty"`
-	StageFixed  bool   `mapstructure:"stage_fixed" yaml:"stage_fixed,omitempty" json:"stage_fixed,omitempty" toml:"stage_fixed,omitempty"`
+	FailText       string `mapstructure:"fail_text"        yaml:"fail_text,omitempty"   json:"fail_text,omitempty"        toml:"fail_text,omitempty"`
+	Interactive    bool   `mapstructure:"interactive"      yaml:",omitempty"            json:"interactive,omitempty"      toml:"interactive,omitempty"`
+	UseStdin       bool   `mapstructure:"use_stdin"        yaml:",omitempty"            json:"use_stdin,omitempty"        toml:"use_stdin,omitempty"`
+	StageFixed     bool   `mapstructure:"stage_fixed"      yaml:"stage_fixed,omitempty" json:"stage_fixed,omitempty"      toml:"stage_fixed,omitempty"`
+	FilesFromStdin bool   `mapstructure:"files_from_stdin" yaml:",omitempty"            json:"files_from_stdin,omitempty" toml:"files_from_stdin,omitempty"`
 }
 
 func (c Command) Validate() error {
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -25,11 +25,10 @@ type Command struct {
 	Exclude  string `mapstructure:"exclude"  yaml:",omitempty" json:"exclude,omitempty"  toml:"exclude,omitempty"`
 	Priority int    `mapstructure:"priority" yaml:",omitempty" json:"priority,omitempty" toml:"priority,omitempty"`
 
-	FailText       string `mapstructure:"fail_text"        yaml:"fail_text,omitempty"   json:"fail_text,omitempty"        toml:"fail_text,omitempty"`
-	Interactive    bool   `mapstructure:"interactive"      yaml:",omitempty"            json:"interactive,omitempty"      toml:"interactive,omitempty"`
-	UseStdin       bool   `mapstructure:"use_stdin"        yaml:",omitempty"            json:"use_stdin,omitempty"        toml:"use_stdin,omitempty"`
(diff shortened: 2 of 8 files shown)


mrexox:
> Lefthook is currently not completely ready for repositories with filenames that contain commas

Could you please share an example repo with such files? Why there's a need to have commas in file names? I am just curious, because I've never met this before.

cmd/run.go
```diff
@@ -36,7 +36,14 @@ func newRunCmd(opts *lefthook.Options) *cobra.Command {
 		"run hooks on all files",
 	)
 
-	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "run on specified files. takes precedence over --all-files")
+	runCmd.Flags().BoolVar(
+		&runArgs.FilesFromStdin, "files-from-stdin", false,
+		"get files from standard input, null-separated, instead of from CLI parameters --file(s)",
+	)
+
+	runCmd.Flags().StringSliceVar(&runArgs.Files, "files", nil, "[DEPRECATED] run on specified files, comma-separated. takes precedence over --all-files")
```
mrexox, line 44: Please, refer this [this code example](https://github.com/evilmartians/lefthook/blob/master/cmd/install.go#L31) to mark deprecations.

internal/config/command.go
```diff
@@ -25,10 +25,11 @@ type Command struct {
...
-	Interactive bool   `mapstructure:"interactive" yaml:",omitempty"            json:"interactive,omitempty" toml:"interactive,omitempty"`
-	UseStdin    bool   `mapstructure:"use_stdin"   yaml:",omitempty"            json:"use_stdin,omitempty"   toml:"use_stdin,omitempty"`
-	StageFixed  bool   `mapstructure:"stage_fixed" yaml:"stage_fixed,omitempty" json:"stage_fixed,omitempty" toml:"stage_fixed,omitempty"`
+	FailText       string `mapstructure:"fail_text"        yaml:"fail_text,omitempty"   json:"fail_text,omitempty"        toml:"fail_text,omitempty"`
+	Interactive    bool   `mapstructure:"interactive"      yaml:",omitempty"            json:"interactive,omitempty"      toml:"interactive,omitempty"`
+	UseStdin       bool   `mapstructure:"use_stdin"        yaml:",omitempty"            json:"use_stdin,omitempty"        toml:"use_stdin,omitempty"`
+	StageFixed     bool   `mapstructure:"stage_fixed"      yaml:"stage_fixed,omitempty" json:"stage_fixed,omitempty"      toml:"stage_fixed,omitempty"`
+	FilesFromStdin bool   `mapstructure:"files_from_stdin" yaml:",omitempty"            json:"files_from_stdin,omitempty" toml:"files_from_stdin,omitempty"`
```
mrexox, line 32: I can't imagine a case when you need this option implicitly, because as I know `git` does not pass files to STDIN. So, I assume the only use case is when you explicitly run something like `echo file1 file2 | lefthook run do-something`. Do I miss something?

internal/lefthook/run/prepare_command.go
```diff
@@ -64,8 +66,35 @@ func (r *Runner) prepareCommand(name string, command *config.Command) (*run, err
...
+	for i, c := range s {
+		if c == 0 {
+			result = append(result, s[start:i])
+			start = i + 1
+		}
+	}
+	result = append(result, s[start:])
+	return result
+}
+
```
mrexox, line 81: I'd like this to be in `internal/lefthook/run.go`, so you parse the files only once and then pass them to all commands as if they were passes via `--files` arg. Otherwise if you have multiple commands using files, only one will receive them.

internal/lefthook/run.go
```diff
@@ -124,6 +125,7 @@ Run 'lefthook install' manually.`,
 			SkipSettings:    logSettings,
 			DisableTTY:      cfg.NoTTY || args.NoTTY,
 			AllFiles:        args.AllFiles,
+			FilesFromStdin:  args.FilesFromStdin,
```
mrexox, line 128: I think there's no need to pass this argument. You can parse the files here and pass them via `Files` option.

sanmai-NL:
What you're asking to me is like, can you show an example of someone needing a macro in Word document sent over e-mail? The problem is that you don't know which files people supply, e.g. in a PR branch. These are valid filenames.

If you're asking for user stories rather than attacker stories, then as a user, I want to script Lefthook in a cross-platform way, without manipulating e.g., Git tool output to replaced null chars or newlines with a comma. Mind that `tr` is a Unix tool.

mrexox:
I am just curious. I agree that it is a possible case. I don't mind adding the support for getting files from STDIN and passing files via `--file` option if you address my comments above.

internal/lefthook/run.go
sanmai-NL: If the functionality isn't behind a CLI option, it will break cases where users supply other information to standard input (e.g., using `yes`).

internal/config/command.go
sanmai-NL: With implicitly you mean, from configuration?

In the use case of Git hooks, then true. But Lefthook can be invoked in multiple ways for multiple use cases. For example, an organization may prohibit acceptance of paths via `--files` (for example, because it does have filenames with comma's in it, like data science CSV batches produced outside of their control) by enforcing this in the team-wide Lefthook config via `files_from_stdin`.

mrexox: With implicitly I mean from the git hook. If you specify `lefthook run do-something --files-from-stdin`, then you probably expect that `{files}` or `{staged_files}` are replaced from the data parsed from STDIN. But there's no need to have a special template for this case.

internal/lefthook/run.go
mrexox: Sorry, I don't get it. If the functionality isn't behind a CLI option, then lefthook is called from within a git hook. In this case files from STDIN don't make sense, since git doesn't pass files via STDIN.

And if lefthook is called explicitly, I'd like to make sure that it is called correctly. If the override is required – let's rely on the CLI options.

I want to distinguish the implicit calls that must make sense and explicit calls that tweak lefthook behavior, for instance change the files passed to the commands. That's my vision and I believe it aligns with KISS principle.

internal/config/command.go
sanmai-NL: Not for that case, indeed. Please note my previous reply though.

internal/lefthook/run.go
sanmai-NL: > Sorry, I don't get it. If the functionality isn't behind a CLI option, then lefthook is called from within a git hook. In this case files from STDIN don't make sense, since git doesn't pass files via STDIN.

First, let's discuss Git hook behavior. According to its docs, Git hook do get information through standard input, see e.g.: https://git-scm.com/docs/githooks#_pre_push

> And if lefthook is called explicitly, I'd like to make sure that it is called correctly. If the override is required – let's rely on the CLI options.

That was exactly my intention, but previously you wrote:
>  I think there's no need to pass this argument.

internal/config/command.go
mrexox: > An organization may prohibit acceptance of paths via --files (for example, because it does have filenames with comma's in it, like data science CSV batches produced outside of their control) by enforcing this in the team-wide Lefthook config via files_from_stdin

I am sorry, I can't agree on this use case because it doesn't align with the main purpose of lefthook – git hooks management. Adding `{files_from_stdin}` template doesn't fit into my vision of what lefthook should support because it adds a very special flow not related to git hooks at all.

I am ok with this change: `--files-from-stdin` option forces lefthook to get the file list from STDIN and replace `{files}` or `{staged_files}` with the files from STDIN.

cmd/run.go
sanmai-NL: Done in [ea12589](https://github.com/evilmartians/lefthook/pull/596/commits/ea125897aa68cbeec62b9b21bf99e633d24931b2).

internal/config/command.go
sanmai-NL: @mrexox Can you check again whether I'm truly adding a template `{files_from_stdin}` and if so, please point out which source code sections do this, so that I can remove them while keeping a configuration option `files_from_stdin`.

internal/lefthook/run/prepare_command.go
sanmai-NL: @mrexox My intention was that you had to choose: either from stdin, or `--file` (or deprecated `--files`). This to reduce confusion. And of course, this behavior should be checked at runtime and documented (marking `--files-from-stdin` and `--file(s)` as mutually exclusive). Do you suggest something else here?

internal/lefthook/run.go
mrexox: > I think there's no need to pass this argument.

It was related to Go code. I meant it's better to read from STDIN in `internal/lefthook/run.go` and pass the results as `Files` option. It's Ok to have `--files-from-stdin` CLI option that replaces files in lefthook commands.

> Git hooks do get information through standard input

Yes, but it is related to Git while getting files from STDIN is not about git.

internal/lefthook/run/prepare_command.go
mrexox: I'm Ok with this vision. We can return an error if both options get passed. But we could also do the following:

In case user calls lefthook this way: `lefthook run some-hook --files file.ts --files-from-stdin
1. Get the files from `--files` option
2. Parse files from STDIN
3. Concat the results and pass to Runner

This is a more inclusive option and I like it too.

sanmai-NL: That is fine, I'm working on it.

sanmai-NL: Done in https://github.com/evilmartians/lefthook/pull/596/commits/c661ed217df8c58ce4d34e1c992c6169dabb6be3

My previous implementation was not completely correct, since it split on end-of-line characters too. But I'm not a Go programmer and misunderstood the semantics. 

I find it strange that Go uses a `string` type for filesystem operations, that may or may not be a grammatically valid filesystem path. And that this string type isn't even a string but a byte array.

internal/config/command.go
sanmai-NL: > > An organization may prohibit acceptance of paths via --files (for example, because it does have filenames with comma's in it, like data science CSV batches produced outside of their control) by enforcing this in the team-wide Lefthook config via files_from_stdin
> 
> I am sorry, I can't agree on this use case because it doesn't align with the main purpose of lefthook – git hooks management. Adding `{files_from_stdin}` template doesn't fit into my vision of what lefthook should support because it adds a very special flow not related to git hooks at all.

Then the above reflects where we stood, in our discussion.

Data science teams do record data files in Git (e.g., with Git LFS), and then these issues with commas pop up. An organizations should be able to enforce `files_from_stdin` and the only thing that can be made constant is an immutable config (file).

[This official guidelines document](https://records-express.blogs.archives.gov/2017/08/22/best-practices-for-file-naming/) warns against using commas in filenames, which means it does happen in some datasets and software should not break (indeed, a linter should be possible that detects this, this is where Lefthook would help).

Anecdotes:
- https://unix.stackexchange.com/a/155836 (search for comma)
- https://forums.autodesk.com/t5/vault-forum/filenames-with-commas/td-p/9464972

sanmai-NL:
@mrexox I'm very keen to get this merged, so that my team has another blocker removed.

Would you be so kind to develop tests and to write a (draft) docs/changelog entry to your liking already, even if one thing still must be ironed out?

internal/lefthook/run.go
mrexox: > Git passes paths via stdin to a hook script, such as Lefthook, in some cases.

Git passes data with a special format: https://git-scm.com/docs/githooks#_pre_push. This is not files list separated with newlines. So, an option `files_from_stdin` can't be safely used for a git hook. In this case this option is only valid for the explicit call of lefthook. This means we can keep just a CLI argument `--files-from-stdin`

mrexox: Let me show how I see the configuration:

```yml
# lefthook.yml

custom-hook:
  commands:
    lint:
      files: git diff --name-only main # custom files
      run: python scripts/linter.py {files}
```

When called like `lefthook run custom-hook`, `{files}` will be replaced with the results of `git diff --name-only main`. When called like `find . -name '*.py' -print0 | lefthook run custom-hook --files-from-stdin` `{files}` will be replaced from the \0-separated list from STDIN.

Does it sound good?

mrexox:
I'll be glad to help with polishing the PR and adding documentation and changelog entries when we finish the discussion :+1:

internal/lefthook/run.go
sanmai-NL: @mrexox Only for pre-push, pre-receive and reference-transaction stdin gets used. I was intending for the configuration option to apply even when Lefthook is invoked from the CLI. I prefer check for those hooks that use stdin in the source code rather than not having the config key.

sanmai-NL:
@mrexox Can we continue this ?

mrexox:
@sanmai-NL, yes, I plan to add a few edits to the PR:

- Instead of passing AllFiles and FilesFromStdin, just get the files once and pass them in a Files option
- Remove `files_from_stdin` config option. I think that it is impossible to use this option with git hooks, so I plan to leave it only as an override for the explicit call like `find . -name '*.go' | lefthook run pre-commit --files-from-stdin`

I can do them myself since I think I've got the vision.

sanmai-NL:
Yes please continue, that sounds acceptable to my personal vision (although as discussed, removing `files_from_stdin` as config option entirely seems a bit unnecessary, but alas). I don't know about the advantages of your first design change, but I suppose you have thought that through.

sanmai-NL:
@mrexox, but splitting on both `\n` and `\0` breaks my intentional functionality ... 

I'm only now ready to review. Regrettably, it seems I assessed your change differently from how it was implemented. I thought you also added an option to split on newlines (exclusively), but you combined it with the null character based splitting.

Splitting on `\n` is not safe, same as on `,`, since anyone can create and supply files with such characters in their names. This choice has the consequence that I can still not pipe Unix tool output (see OP) since sometimes, newlines are in there too even without a file name containing one. The freedom to combine unix tool outputs, in addition to safety, is why all other tools do not split on newlines ...

mrexox:
@sanmai-NL, having newlines in file names is not conventional (however is possible). Do you have such files, or are you afraid that it can break something for your work?

sanmai-NL:
The latter, such files can be added to pull request branches and it would break tooling.

sanmai-NL:
Or rather, it would break Lefthook but not other tooling.

mrexox:
Alright, I will change this in next fix version to use only `\0` by default

PR #606, opened 2024-01-12
fix: safe execute git commands without sh wrapper

mrexox:
**:wrench: Summary**


sanmai-NL:
Are strings embedded in shell commands now sanitized in the remaining case? You can use this if not https://pkg.go.dev/github.com/alessio/shellescape

Issue #292, opened 2022-06-29
Run `files` cmd with `root` parameter


KubaJastrz:
Files array is computed by running a `files` command from working directory. But `run` is run from `filepath.Join(r.repo.RootPath, command.Root)`. This makes a weird looking configuration in case of using custom scripts for both options:

```yml
# consider nested packages/test/lefthook.yml
pre-push:
  commands:
    eslint:
      root: 'packages/test/'
      files: scripts/get-files-push               # this is relative to working directory (repo root)
      run: ../../scripts/lint-fix eslint {files}  # this is relative to packages/test
```

In my opinion `files` should also be computed based on `root` path.

Source code:

https://github.com/evilmartians/lefthook/blob/f3ffc5bc5bc89e5646d6eaad570eefc27938c471/internal/git/repository.go#L88-L93

https://github.com/evilmartians/lefthook/blob/3e6c8eca9f0c9b899244ebbc23ef37d161a7118f/internal/lefthook/runner/runner.go#L191-L192

jameslporter:
Love lefthook!! I just dove in a day or so ago. This was the only real pain point for me. We have a CRA app in client directory and server node code in the root level folders. Below is the run command I came up with to fix the paths by removing the root level of the path. If someone needs this but deeper into the directory structure adjust the cut -d'/' -f2- part to make the number by the f be +1 of how many directories to lop off.

`run: fixedFiles=$(for staged_file in ${staged_files}; do echo $(echo $staged_file | cut -d'/' -f2-); done;); cd client; prettier --ignore-unknown --write $fixedFiles && eslint --max-warnings 0 $fixedFiles && git add $fixedFiles`

Unrelated but thought I'd share this idea while I'm here. I made a post-merge action that globs on lefthook.yml to trigger lefthook install. This way new developers work out of the box after npm i from a fresh clone and existing developers will be kept up to date as the config evolves.

```
post-merge:
  commands: 
    lefthook:
      glob: "lefthook.yml"
      run: npm exec lefthook install
```

only other minor feedback is occasionally it spits characters out into my terminal after running npm exec lefthook run pre-commit. Not very consistent so repro is difficult of course. Anyways, thanks for this contribution and plus one to this feature request.

PR #607, opened 2024-01-12
fix: execute files command within configured root

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/292

**:wrench: Summary**

Execute `files` from within a configured root in case it contains a path to a script. This is a breaking change.


jameslporter:
Are the files passed in staged_files corrected to be relative to the defined root path? That's my pain point, had to make a gnarly shell script to chop client/ off the beginning of our paths.

mrexox:
@jameslporter , yes, files are being filtered by `root` option: https://github.com/evilmartians/lefthook/blob/master/internal/lefthook/run/filter/filters.go#L22

PR #634, opened 2024-02-21
handle "run" command in the skip/only setting

### LINKED ISSUE #598: Using conditions for Skip and Only
### :zap: Summary

Implementing conditional support for the `skip` and `only` options would improve control over hook execution and performance by skipping unnecessary steps and providing a flexible solution to meet different project requirements.

### Value

Implementing a feature to use conditions for the `skip` and `only` options can provide several benefits:

- **Selective Execution:** Conditions allow developers to selectively execute hooks based on specific criteria, enabling more granular control over when scripts run. This can be useful for running scripts only on certain branches, commit messages, or file changes.
- **Optimized Workflow:** With conditional execution, you can optimize the workflow by skipping unnecessary steps in certain scenarios. For instance, you might want to skip certain checks on feature branches or only run specific tasks on the main branch.
- **Customizable Automation:** Conditions provide a way to customize automation based on project requirements. This flexibility allows teams to tailor their hook scripts to fit the unique needs of their development process.
- **Shared Configurations**: makes it easier to integrate with shared remote configurations that could be more generic and customizable.
- **Enhanced Code Quality:** By selectively applying hooks based on conditions such as file paths or commit messages, you can enforce coding standards and perform checks relevant to the changes being made. This contributes to maintaining and enhancing overall code quality.
- **Improved Performance:** Skipping unnecessary hooks can improve performance by reducing the execution time of CI/CD processes. This is crucial for maintaining efficient and speedy development pipelines.

Incorporating these features enhances the adaptability and efficiency of Lefthook, empowering teams to build a more tailored and effective continuous integration environment.

Below is a small example of how the configuration could look using this new feature

```yaml
pre-commit:
  commands:
    sast:
      only:
        - run: test -n "${USER}" && test -n "${PASS}"
      glob: "*.go"
      run: my_sast_tool --user "${USER}" --pass "${PASS}" {staged_files}
    dast:
      skip:
        - run: test "${NO_DAST}" -eq 1
        - run: npx is-ci
        - ref: dev/*
      glob: "*.go"
      run: my_dast_tool {staged_files}
```

prog-supdex:
Closes #598 

#### :zap: Summary

Added the ability to handle "run" commands in "skip" and "only" sections.

There are a few restrictions

1. We can only put one "run" command in `skip`/`only` sections because `skip/only` parses like a `mapstructure`. So we can't write
skip
        - run: test "${NO_DAST}" -eq 1
        - run: npx is-ci
        - ref: dev/*
 `
 
2. In this PR, we can't put the command with pipelines (`&&, &, ||` etc.), like this `test "${NO_DAST}" -eq 1 && test ....`, because this command is executed via `exec.Command(cmdName, cmdArgs...)`, which accepts the first argument as the command and the rest of the argument as the command arguments.

3. Since `only` is the opposite of `skip,` if the `run` command returns `true` for the `skip` section, it will skip, but for the `only` section, it must return `false.`

- [x] Check locally
- [x] Add tests
- [x] Add documentation


internal/config/skip.go
```diff
@@ -4,47 +4,90 @@ import (
...
+
+func NewSkipChecker(executor Exec) *SkipChecker {
+	if executor == nil {
+		executor = NewOsExec()
+	}
+
+	return &SkipChecker{Executor: executor}
+}
+
+func (sc *SkipChecker) DoSkip(gitState git.State, skip interface{}, only interface{}) bool {
```
mrexox, line 22: Let's change the name of this method? I think for the `SkipChecker` it's more intuitive to use `Check` method. WDYT?

internal/config/skip.go
```diff
@@ -4,47 +4,90 @@ import (
...
+	return g.Match(gitState.Branch)
+}
+
+func (sc *SkipChecker) matchesCommands(typedState map[string]interface{}) bool {
+	commandLine, ok := typedState["run"].(string)
+	if !ok {
+		return false
+	}
+
+	log.Debug("[lefthook] skip/only cmd: ", commandLine)
```
mrexox, line 90: It would be great to log the exit code here (or the result of a check)

internal/config/script.go
```diff
@@ -24,7 +24,8 @@ type Script struct {
 }
 
 func (s Script) DoSkip(gitState git.State) bool {
-	return doSkip(gitState, s.Skip, s.Only)
+	skipChecker := NewSkipChecker(nil)
```
mrexox, line 27: I don't like this passing `nil`. Could we pass `NewOsExec()` explicitly?

internal/config/skip.go
mrexox, line 1: We can name this file skip_checker.go since now it contains a SkipChecker struct

mrexox (commented):
I like the results. It would be nice to have integrity tests (within `testdata/` folder) but it looks great.

A few nitpicks related to naming and argument passing.

internal/config/skip.go
prog-supdex: Done

internal/config/skip.go
prog-supdex: Done

mrexox (changes requested):
I've tested the PR with the following config:

```yml
skip:
  skip:
    - run: test -z $VAR
  commands:
    echo:
      run: echo "should not be printed - $VAR"
```

```bash
VAR=1 lefthook run skip -v
```

But it always prints VAR whenever it is set or not.

I think the problem is that the arguments are passes as raw args to a cmd. Maybe it makes sense to wrap the commandLine with `sh -c` the way we do it in `exec_unix.go`. 

The only concern I have regarding this is that on Windows the feature may not work. But anyway $VAR is sh syntax, not a system syntax, so it makes sense to me.

internal/config/exec.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Cmd runs plain string command. It checks only exit code and returns bool value.
+func (o *osExec) Cmd(commandLine string) bool {
+	if commandLine == "" {
+		return false
+	}
+
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
```
prog-supdex, line 26: @mrexox I added calling the command via `sh -c` but as you said, it may not work with `Windows` if it does not include WSL, Cygwin, and so on.

So I added checking the OS, but I couldn't check it because we don't have a running test-integrity for windows [here](https://github.com/evilmartians/lefthook/blob/master/.github/workflows/test.yml#L35)

I think there was a reason why it was not added to ci-cd for windows :)
But what do you think about checking the `windows` and adding test integrity for Windows to ci-cd?

mrexox: Let's try it out!

prog-supdex: @mrexox, Unfortunately, the `[windows] skip' command was necessary to skip some of the Windows test integrations.

Also, I use `PowerShell` for Windows commands because it supports more complex commands (e.g. with `if`).

We can avoid skipping some `test-integrities`, but it requires adapting those tests, and it is difficult without a Windows machine.

The example of skipping test [here](https://github.com/evilmartians/lefthook/pull/634/files#diff-3d168f7bd5c5e2a73b3bc1a841945beac68cb9960ba566d4716428a845c99981R1)

mrexox: Alright, I think this is acceptable for now. Thank you for digging into testscript, I appreciate it. I will try to figure out how to make it work for Windows when I have a chance to work on a Windows machine.

mrexox (approved):

PR #637, opened 2024-02-23
Add new output instead skip_output

### LINKED ISSUE #592: Not just `skip_output`, specify `output` explicitly
### :zap: Summary

Currently, it's only possible to further reduce output rather than increase it.

### Value

I want to see which commands were executed, e.g., `execution_info` to help my users and to debug issues.

Deprecate `skip_output` and change it to `output` which is an explicit positive/allowlist of outputs.

prog-supdex:
Closes #592 

#### :zap: Summary

Deprecate the `skip_output` and add `output` as a whitelist of output values
I saved the `skip_output` logic for further removal in the next versions

- [x] Check locally
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/internal/lefthook/run.go
+++ b/internal/lefthook/run.go
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	envEnabled    = "LEFTHOOK"       // "0", "false"
-	envSkipOutput = "LEFTHOOK_QUIET" // "meta,success,failure,summary,skips,execution,execution_out,execution_info"
+	envEnabled    = "LEFTHOOK"        // "0", "false"
+	envSkipOutput = "LEFTHOOK_QUIET"  // "meta,success,failure,summary,skips,execution,execution_out,execution_info"
+	envOutput     = "LEFTHOOK_OUTPUT" // "meta,success,failure,summary,skips,execution,execution_out,execution_info"
 )
 
 type RunArgs struct {
@@ -75,10 +76,18 @@ func (l *Lefthook) Run(hookName string, args RunArgs, gitArgs []string) error {
 		log.SetLevel(log.WarnLevel)
 	}
 
+	newTags := os.Getenv(envOutput)
 	tags := os.Getenv(envSkipOutput)
 
-	var logSettings log.SkipSettings
-	(&logSettings).ApplySettings(tags, cfg.SkipOutput)
+	var logSettings log.SettingsInterface
+
+	if tags == "" && cfg.SkipOutput == nil {
+		logSettings = log.NewSettings()
+		logSettings.ApplySettings(newTags, cfg.Output)
+	} else {
+		logSettings = log.NewSkipSettings() //nolint:staticcheck //SA1019: for temporary backward compatibility
+		logSettings.ApplySettings(tags, cfg.SkipOutput)
+	}
 
 	if !logSettings.SkipMeta() {
 		log.Box(
@@ -192,7 +201,7 @@ Run 'lefthook install' manually.`,
 func printSummary(
 	duration time.Duration,
 	results []run.Result,
-	logSettings log.SkipSettings,
+	logSettings log.SettingsInterface,
 ) {
 	summaryPrint := log.Separate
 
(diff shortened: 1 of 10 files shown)


internal/log/log.go
```diff
@@ -60,6 +60,19 @@ const (
 	spinnerText        = " waiting"
 )
 
+type SettingsInterface interface {
```
mrexox, line 63: I think we should name it just `Settings`

internal/log/log.go
```diff
@@ -60,6 +60,19 @@ const (
...
+	ApplySettings(tags string, skipOutput interface{})
+	SkipSuccess() bool
+	SkipFailure() bool
+	SkipSummary() bool
+	SkipMeta() bool
+	SkipExecution() bool
+	SkipExecutionOutput() bool
+	SkipExecutionInfo() bool
+	SkipSkips() bool
+	SkipEmptySummary() bool
```
mrexox, line 73: Since we move away from negation (`output` instead of `skip_output` and `Settings` instead of `SkipSettings`) I think it makes sense to change these methods to `LogSuccess(), LogFailure(), ...`. It sounds more natural, doesn't it?

WDYT?

internal/log/log.go
mrexox: Let's move this interface to the `settings.go`?

internal/log/settings.go
```diff
@@ -0,0 +1,124 @@
...
+	summary
+	skips
+	execution
+	executionOutput
+	executionInfo
+	emptySummary
+	enableAll = ^0 // Set all bits as 1
+)
+
+type Settings int16
```
mrexox, line 20: We can name it as `OutputSettings` if we rename the `SettingsInterface` to just `Settings`

mrexox (commented):
I've left a few nitpicks about naming. But the code looks good to me!

internal/lefthook/run.go
```diff
@@ -75,10 +76,18 @@ func (l *Lefthook) Run(hookName string, args RunArgs, gitArgs []string) error {
...
 	}
 
+	newTags := os.Getenv(envOutput)
 	tags := os.Getenv(envSkipOutput)
 
-	var logSettings log.SkipSettings
-	(&logSettings).ApplySettings(tags, cfg.SkipOutput)
+	var logSettings log.SettingsInterface
+
+	if tags == "" && cfg.SkipOutput == nil {
```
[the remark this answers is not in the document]
prog-supdex: oh, `newTags` was a temporary name, I just forgot about it and didn`t see it :)

Thanks, I fixed it!

internal/lefthook/run.go
```diff
@@ -75,10 +76,18 @@ func (l *Lefthook) Run(hookName string, args RunArgs, gitArgs []string) error {
...
+		logSettings = log.NewSettings()
+		logSettings.ApplySettings(newTags, cfg.Output)
+	} else {
+		logSettings = log.NewSkipSettings() //nolint:staticcheck //SA1019: for temporary backward compatibility
+		logSettings.ApplySettings(tags, cfg.SkipOutput)
```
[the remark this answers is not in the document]
prog-supdex: Done

internal/log/log.go
prog-supdex: done

internal/log/settings.go
prog-supdex: done

internal/log/log.go
prog-supdex: yes, I agree, but I kept the same methods for compatibility with the `SkipSettings` and to avoid changing `Skip*` methods everywhere in the `Lefthook` code and I did not want to change other places :)

If I change methods in `Settings`, it means I need to change methods in `SkipSettings` and change it in other places, but if it is ok if I rename `Skip*` methods everywhere, no problem I will do it! :)

mrexox: I think it makes sense to change the naming so it is more intuitive :+1:

prog-supdex: done!

mrexox (approved):
Awesome work! Thank you!

Issue #588, opened 2023-11-27
lefthook can swallow stdin from user typing at the terminal unexpectedly


technicalpickles:
### :wrench: Summary

consider a workflow like this:

```
$ git commit -m "She sells seashells by the sea shore"
<lefthook happens here>
$ git push
$ git checkout
```

Out of muscle memory, it's really easy to hit enter, and immediately start typing the next command while the first command is still running. It seems that lefthook ends up receiving that command you start typing though, so when the first command is finished, the input with the second command doesn't get to the shell.

For some additional context, I've gotten this complaint from a few people after migrating from [committer](https://github.com/gusto/committer) to lefthook.

### Lefthook version

1.5.2

### Steps to reproduce

`lefthook.yaml`:

```yaml
pre-commit:
  commands:
    sleep:
      run: sleep 2
```

- make some change
- type `git commit -a -m "test"<enter>`
- while lefthook is running, type: `git checkout main<enter>`

### Expected results

shell receives the input for the next command, ie `git checkout main<enter>`

### Actual results

input with second command, `git checkout main`, isn't received by the shell.

### Possible Solution

I think it should be possible to close stdin, so it doesn't receive any of the input from a user. I don't know if there are any cases where lefthook does read from stdin though.

### Logs / Screenshots

https://github.com/evilmartians/lefthook/assets/159/ed77fd94-9e1f-4a95-bf53-4292572d052d

noaelad:
I also have this issue, been pretty annoyed with lefthook as a result

att14:
Even with `no_tty` and a full `skip_output` I observe this issue.

```yml
no_tty: true

skip_output:
  - meta           # Skips lefthook version printing
  - summary        # Skips summary block (successful and failed steps) printing
  - empty_summary  # Skips summary heading when there are no steps to run
  - success        # Skips successful steps printing
  - failure        # Skips failed steps printing
  - execution      # Skips printing any execution logs (but prints if the execution failed)
  - execution_out  # Skips printing execution output (but still prints failed commands output)
  - execution_info # Skips printing `EXECUTE > ...` logging
  - skips          # Skips "skip" printing (i.e. no files matched)
  - meta           # Skips lefthook version printing

post-checkout:
  commands:
    sleep:
      run: sleep 2
```

```sh
$ git checkout
M       lefthook.yml
Your branch is up to date with 'origin/main'.
asdf asdf asdf asdf%
$ sdf asdf asdf
```

[...]

PR #638, opened 2024-02-23
Fix lefthook capturing stdin from terminal even if when non-interactive and not using stdin

technicalpickles:
Closes  https://github.com/evilmartians/lefthook/issues/588

#### :zap: Summary

If I'm reading the code, this `args.in` should be a nullReader when non-interactive and not using stdin. For some reason though, it ends up capturing stdin. That is, any typing you do while lefthook is running goes to lefthook, instead of the shell invoking it.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/lefthook/run/exec/execute_unix.go
+++ b/internal/lefthook/run/exec/execute_unix.go
@@ -102,8 +102,6 @@ func (e CommandExecutor) execute(ctx context.Context, cmdstr string, args *execu
 
 		defer func() { _ = p.Close() }()
 
-		go func() { _, _ = io.Copy(p, args.in) }()
-
 		_, _ = io.Copy(args.out, p)
 	}
 


technicalpickles:
>  Add tests

Would love to have tests for this, but I honestly have no idea about how to do it 😬 specifically because I have yet to reproduce https://github.com/evilmartians/lefthook/issues/588 in an automated way

technicalpickles:
I'm not sure why coveralls would fail 🤔 if anything, there's one less line to cover right?

CleanShot 2024-02-23 at 12 13 27

mrexox:
That's OK. I think there are no tests that can cover this. Actually this part is legacy since it was here before the `use_stdin` and `interactive` options. So, it probably doesn't make sense.

I will have this PR tested locally and then will come back with the resolution.

Thank you for this finding!

PR #649, opened 2024-02-27
Update lipgloss

technicalpickles:
Closes https://github.com/evilmartians/lefthook/issues/588

#### :zap: Summary

I don't understand how it could possibly be messing with stdin, but it seems that lipgloss 0.9.1, which is used to output nicely colored logs, causes stdin to be captured.

0.9.1 was last released last October, and there's been updates since
  then. On a whim, I tried the latest commit, and I am shocked to say it
  does *not* swallow stdin.

I'm truly perplexed.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.21
 require (
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/briandowns/spinner v1.23.0
-	github.com/charmbracelet/lipgloss v0.9.1
+	github.com/charmbracelet/lipgloss v0.9.2-0.20240213153121-13584f26deeb
 	github.com/creack/pty v1.1.18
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/muesli/reflow v0.3.0 // indirect
 	github.com/muesli/termenv v0.15.2 // indirect
 	github.com/pelletier/go-toml/v2 v2.1.1
-	github.com/rivo/uniseg v0.2.0 // indirect
+	github.com/rivo/uniseg v0.4.6 // indirect
```
technicalpickles, line 39: I guess it's plausibly this dependency 🤔

technicalpickles: IT IS THE DEPENDENCY

PR #650, opened 2024-02-27
Update uniseg

technicalpickles:
[Closes # (issue)](https://github.com/evilmartians/lefthook/issues/588)

#### :zap: Summary

I don't understand how it could possibly be messing with stdin, but uniseg 0.2.0, which is a dependency of lipgloss, which we use for nicely colored logs... is causing stdin to be captured.

Originally, I [thought it was lipgloss](https://github.com/evilmartians/lefthook/pull/649/), but there weren't any a [huge diff](https://github.com/charmbracelet/lipgloss/compare/v0.9.1...master), and nothing immediately jumped out. I did notice uniseg is a dependency, and was also updated.

So I upgraded it on a whim, and it no longer swallows stdin 🙃 

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/muesli/reflow v0.3.0 // indirect
 	github.com/muesli/termenv v0.15.2 // indirect
 	github.com/pelletier/go-toml/v2 v2.1.1
-	github.com/rivo/uniseg v0.2.0 // indirect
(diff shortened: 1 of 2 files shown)


mrexox:
Well, that's mysterious! But this is a great catch. Thank you for digging into it!

PR #653, opened 2024-02-28
Configurable path to lefthook to run during git hook

technicalpickles:
#### :zap: Summary

This adds `LEFTHOOK_PATH` to the `.git/hook` templates. It's used first before 

This is something that would have been helpful while working on https://github.com/evilmartians/lefthook/issues/588 and I was testing against a locally built 

While I was here, I added `set -x` when `LEFTHOOK_VERBOSE` is on. This is another thing that would have been useful, and I also used it to help validate LEFTHOOK_PATH working 😁  Let me know if I should split it out separately.

- [x] Check locally
- [ ] Add tests
- [x] Add documentation

### DIFF
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -19,9 +19,9 @@ call_lefthook()
   osArch=$(uname | tr '[:upper:]' '[:lower:]')
   cpuArch=$(uname -m | sed 's/aarch64/arm64/')
 
-  if test -n "$LEFTHOOK_PATH"
+  if test -n "$LEFTHOOK_BIN"
   then
-    "$LEFTHOOK_PATH" "$@"
+    "$LEFTHOOK_BIN" "$@"
   elif lefthook{{.Extension}} -h >/dev/null 2>&1
   then
     lefthook{{.Extension}} "$@"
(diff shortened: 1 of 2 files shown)


internal/templates/hook.tmpl
```diff
@@ -15,7 +19,10 @@ call_lefthook()
   osArch=$(uname | tr '[:upper:]' '[:lower:]')
   cpuArch=$(uname -m | sed 's/aarch64/arm64/')
 
-  if lefthook{{.Extension}} -h >/dev/null 2>&1
+  if test -n "$LEFTHOOK_PATH"
+  then
+    "$LEFTHOOK_PATH" "$@"
```
mrexox, line 24: Hey! Thank you for this PR, it makes sense to add this feature. I think it's not obvious that LEFTHOOK_PATH is not a path to a folder containing `lefthook` executable, but a full path to the lefthook executable. 

I wonder if there's a better name for this feature 🤔 WDYT about `LEFTHOOK_BIN` or `LEFTHOOK_EXECUTABLE`?

technicalpickles: I went with `_BIN`, on my very scientific survey (looking at neovide, which I happened to have open)

mrexox (approved):

mrexox:
Thank you!

Issue #718, opened 2024-05-13
Error message when lefthook is older could include more info


technicalpickles:
### :zap: Summary

If you have a `min_version` in lefthook.yml, and try to run a hook using it, you get a failure like:

```
Error: required Lefthook version is higher than current
```

I'd propose adding some info like:

- the current version running
- the version requested
- the path of lefthook that is running

### Value

This would make it easier to help someone figure out why it's failing.

One example I have run into is with having both homebrew and gem installed lefthook... if you use `rbenv` or another shim based ruby version manager, you typically would have that on the PATH before a homebrew installed one, and it's really easy for those to get out of sync.

mrexox:
That's a good improvement! Thank you. I will try to achieve this but your PR is also welcome.

PR #721, opened 2024-05-17
fix: verbose error on versions mismatch

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/718

**:wrench: Summary**

Now when lefthook version mismatches the following error will be printed:

```
Error: required lefthook version (1.6.13) is higher than current (1.6.11) at /Users/me/go/bin/lefthook
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -3,6 +3,7 @@ package version
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


technicalpickles:
Thanks for this! 🙇🏻

PR #731, opened 2024-05-27
fix: LFS pre-push requires Stdin

### LINKED ISSUE #511: Git LFS doesn't appear to be working
### :wrench: Summary

We have lfs setup on our environments and after installing the lefthook hooks it replaces the git lfs hooks.  From the documentation it lefthook does support lfs and if I turn on verbose I can see it try to execute lfs but not sure if there is secret sauce in the lfs hooks but every push doens't include the lfs file upload.

I have tried both wsl2  and windows and both behave the same.  If I remove the lefthook pre-push hooks and re-instate the lfs ones using `git lfs update --force`.  If i push then i can see that the lfs update has triggered.

The only thing I can think of is that potentially would be different execution environment between the two.  I noticed that the hook uses `git-lfs` and the lefthook call uses `git lfs`.  Lastly I also noticed in the hook for windows it does a `unset GIT_PREFIX; ssh git@bitbucket.org 'git-receive-pack '\''brian_garvey/lfs-lefthook-test.git'\'''`

Here is the test repo [LFS Lefthook Test](https://bitbucket.org/brian_garvey/lfs-lefthook-test/src/master/).  I tried to keep it simple.

### Lefthook version

lefthook -> 1.4.3
git -> 2.41.0
git-lfs -> 3.0.2 (WSL 2) & 3.3.0 (windows)

### Steps to reproduce

1. Copy JPG File to directory
2. Install the lefthook hooks `lefthook install -a`
3. Run `git add .` and `git commit -m 'message'`
4. Rung `git push`

### Expected results

LFS images are pushed to the remote repository and and subsequently pull them down

### Actual results

Check bitbucket UI and see that images are not loading and external pull shows 404 errors for the lfs files

### Logs / Screenshots

git hook version
```
trace git-lfs: pure SSH protocol connection failed: Unable to negotiate version with remote side (unable to read capabilities): EOF
trace git-lfs: pre-push: refs/heads/master a92e62630365e8950c18a69b83ab8c141f97a4d8 refs/heads/master eee5ea60a973bde0729800020680bfd928d6be24
trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'show-ref'
trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'ls-remote' '--heads' '--tags' '-q' 'origin'
trace git-lfs: tq: running as batched queue, batch size of 100
````

lefthook version
```
trace git-lfs: pure SSH protocol connection failed: Unable to negotiate version with remote side (unable to read capabilities): EOF
trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'show-ref'
trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'ls-remote' '--heads' '--tags' '-q' 'origin'
trace git-lfs: filepathfilter: creating pattern ".git" of type gitignore
```

tdesveaux:
DRAFT: Opening as draft. I'll self-review tomorrow and look if I can add tests or update doc

Partial workaround fix for #511

#### :zap: Summary

`git lfs pre-push` hook read information on what will be pushed to remote to determine what LFS object it needs to upload.
Until now, Stdin from Git was not forwarded to it, causing it to never uploading LFS objects, which is a big issue as these objects can be lost due to pruning.

There is another change that check if multiple commands have the `use_stdin` option to true, as, at the moment, Stdin will be consumed by the first command executed, and others will be left hanging.

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1431,6 +1431,10 @@ When you try to commit `git commit -m "bad commit text"` script `template_checke
 > **Note**
 >
 > With many commands or scripts having `use_stdin: true`, only one will receive the data. The others will have nothing. If you need to pass the data from stdin to every command or script, please, submit a [feature request](https://github.com/evilmartians/lefthook/issues/new?assignees=&labels=feature+request&projects=&template=feature_request.md).
+>
+> The LFS `pre-push` hook needs the data provided by Git through Stdin. As such, no other `pre-push` command or script should use this flag when the repository uses LFS.
+>
+> When run, lefthook will error and stop execution if multiple commands or scripts have this flag enabled.
 
 Pass the stdin from the OS to the command/script.
 
--- a/internal/git/lfs.go
+++ b/internal/git/lfs.go
@@ -16,6 +16,10 @@ var lfsHooks = [...]string{
 	"pre-push",
 }
 
+var lfsHookConsumeStdin = [...]string{
+	"pre-push",
+}
+
 // IsLFSAvailable returns 'true' if git-lfs is installed.
 func IsLFSAvailable() bool {
 	_, err := exec.LookPath("git-lfs")
@@ -26,6 +30,18 @@ func IsLFSAvailable() bool {
 // IsLFSHook returns whether the hookName is supported by Git LFS.
 func IsLFSHook(hookName string) bool {
 	for _, lfsHookName := range lfsHooks {
+		if lfsHookName == hookName {
+			return true
+		}
+	}
+
+	return false
+}
+
+// DoesLFSHookConsumeStdin returns whether the LFS hookName will consume Stdin
+// meaning it won't be available to following commands
+func DoesLFSHookConsumeStdin(hookName string) bool {
+	for _, lfsHookName := range lfsHookConsumeStdin {
 		if lfsHookName == hookName {
 			return true
 		}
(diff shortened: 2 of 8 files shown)


tdesveaux:
@mrexox Sorry, I won't be able to spend more time on this right now, and I don't know when I'll have more time.


I'm frankly not satisfied with how the sanity check is implemented right in the middle of run, but I didn't really know where it could be better setup since we don't know if an LFS command will run before that.
I don't really know how to test this, so I'd guidance.

I also changed the fact that LFS errors are ignored. They now stop the hook with an error code. LFS is far too important when used in a repository to ignore.

Since I'm uncertain on when I'd be available to complete this, I don't mind if any one want to take over.

tdesveaux:
taken over in #732

Issue #611, opened 2024-01-16
Lefthook causes git lfs can't store.


Demi871023:
### :wrench: Summary

Lefthook is causing Git LFS failures. Upon pushing a commit that includes LFS files such as jpg, my local repository exhibits no issues. However, when a colleague attempts to pull the same commit, Git LFS fails to locate the associated LFS files, resulting in an error.

### Lefthook version

```
1.5.5
```

### Steps to reproduce

1. Add jpg files to the project.
2. Confirm that the file names are included in the Git LFS tracking by executing the command git lfs ls-files.
3. Utilize lefthook for pushing the code.
4. Encounter a file not found error when others attempt to pull the code.

### Expected results

Neither push code nor pull code will cause errors.

### Actual results

Below is the error message we encountered.
```
git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree checkout -b ycv_1.38_effect_filter --track origin/ycv_1.38_effect_filter 
Downloading assets/effect_preload/effects/mellow_fade/Mellow_Fade_02.jpg (25 KB)
Error downloading object: assets/effect_preload/effects/mellow_fade/Mellow_Fade_02.jpg (5863113): Smudge error: error opening media file: open /Users/xxxx/projectname/.git/lfs/objects/58/63/5863113071a3f649ce8c6774c68bdc556e4e84b2db19c770bcd06730ca588231: no such file or directory

Errors logged to '/Users/xxxx/projectname/.git/lfs/logs/20240115T104854.37545.log'.
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: assets/effect_preload/effects/mellow_fade/Mellow_Fade_02.jpg: smudge filter lfs failed

```

### Possible Solution

We think it may have nothing to do with the file name, because changing the name of Mellow_Fade_02 still cannot result the problem.
At present, we are not sure how to improve the problem when using lefthook, so we can temporarily remove lefthook and then push and pull can be done smoothly without encountering lfs problems.

### Logs / Screenshots

```bash
assert_lefthook_installed: true
pre-push:
  piped: true
  commands:
    linter:
      run: flutter analyze
#    tests:
#      run: flutter test

pre-commit:
  piped: true
  commands:
    format_code:
      glob: '*.dart'
      run: dart format {staged_files} && git add {staged_files}

post-merge:
  commands:
[...]

Issue #147, opened 2020-08-19
pre-push hook missing stdin values


ThisIsMissEm:
According to the git hooks documentation, pre-push receives a line of input from stdin, which gives the branch names:

> This hook is called by git-push[1] and can be used to prevent a push from taking place. The hook is called with two parameters which provide the name and location of the destination remote, if a named remote is not being used both values will be the same.
> Information about what is to be pushed is provided on the hook’s standard input with lines of the form:
> ```
> <local ref> SP <local sha1> SP <remote ref> SP <remote sha1> LF
> ```
> For instance, if the command git push origin master:foreign were run the hook would receive a line like the following:
> ```
> refs/heads/master 67890 refs/heads/foreign 12345
> ```

However, the script that lefthook uses for executing pre-push doesn't forward this input through to the script.

I'm not sure what the right way to fix this is?

ThisIsMissEm:
So, I managed to get this working with the following git hook:

```
#!/bin/sh

if [ "$LEFTHOOK" = "0" ]; then
  exit 0
fi

debug () {
  if [ "$LEFTHOOK_DEBUG" = "true" ] || [ "$LEFTHOOK_DEBUG" = "1" ]; then
    echo "lefthook:debug $1"
  fi
}

hookName="$(basename "$0")"

cmd="lefthook run $hookName $@"

case $hookName in
  "pre-push"|"post-rewrite")
    stdin="$(cat)";;
esac

debug "$hookName"
debug "stdin: $stdin"
debug "cmd: $cmd"

if [ -t 1 ] ; then
  exec < /dev/tty ; # <- enables interactive shell
fi

if lefthook -h >/dev/null 2>&1
then
  debug "eval"
  eval $cmd <<< $stdin
elif bundle exec lefthook -h >/dev/null 2>&1
then
  debug "bundle"
  bundle exec $cmd <<< $stdin
elif npx lefthook -h >/dev/null 2>&1
then
  debug "npx"
  npx $cmd <<< $stdin
elif yarn lefthook -h >/dev/null 2>&1
then
  debug "yarn"
  yarn $cmd <<< $stdin
else
  echo "Can't find lefthook in PATH"
fi
```

Note: rather than outputting a templated file with the hook name as the `.git/hooks/<name>` file, you can just grab the hook name from the file name that is currently being executed; that's what I've done above.

There's probably a better way to solve this.

ThisIsMissEm:
[...]

PR #732, opened 2024-05-29
fix: share STDIN across different commands on pre-push hook

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/511
Closes https://github.com/evilmartians/lefthook/issues/611
Closes https://github.com/evilmartians/lefthook/issues/147

#### :zap: Summary

Right now git lfs pre-push hook does not receive STDIN. We should pass it there but some other hooks may also use it.

For this purpose I've added a `cachedReader` that reads from STDIN once and cached the read data. Not only for `pre-push` but for all hooks that specify `use_stdin` option the `cachedReader` will be used.

However with `parallel: true` the read data may be incorrect because of random reads from different commands that run cuncurrently. So, when you use `use_stdin` for more that one command/script, please consider not using `parallel: true`.

- [x] Check locally
- [x] Add tests
- [ ] Add documentation


mrexox:
Hey @tdesveaux, I've implemented a slightly different approach but it should solve the issue.

1. STDIN is passed to `git lfs` hooks whether it uses it or not. I think this is better to provide it always and avoid being responsible for managin STDIN.
2. But the STDIN is wrapped in a struct that caches all read bytes and allows to re-read them again and again. This solves issues when you have your own script/command that utilizes stdin together with Git LFS.

Do you have some time to test this code for your case? I've done some testing locally but I'd like to hear some feedback from you since this fix should solve your issue.

tdesveaux:
This is a far better approach, thanks for the fix.
I'll test try to test this today.

From a cursory review of the code:
- You should consider writing the buffered Stdin to a file if it's size is large (pretty edge case, but a big push would receive lots of Stdin). I think this can be an noted improvement for later.
- You removed the error escalation if LFS command fails. I'd like to re-iterate how dangerous this is IMHO.

mrexox:
Thank you for the review. Yes, it seems like Git LFS error is a crucial thing so it should fail the whole hook. 

I will consider using file for caching. I think it makes sense when STDIN has about 1+ mb of data. Do you think such cases happen often?

tdesveaux:
https://git-scm.com/docs/githooks#_pre_push

I might have been far too cautious in thinking file buffering might be needed.
There is one line per ref pushed (branch, tag, etc).
For it to attain >1mb, it would take a very large repository being pushed in it's entirety.
Not sure someone doing this would have lefthook enabled.

You can disregard this IMHO, this would add complexity to the code for no real benefit.

tdesveaux:
@mrexox  Just tested this branch. This fix the issue I reproduced.

FYI, here is what I did to reproduce.
- `git init` a repository
- setup lefthook with a `pre-push` hook. (tested with a simple `run: echo pre-push_hook` initially)
- add a dummy `test.bin` file with `echo test >> test.bin`
- `git lfs track test.bin`
- commit `lefthook.yml` and `.gitattributes`
- commit `test.bin`
- push

I also tested your branch with a cheeky:
```
pre-push:
  commands:
    cat-stdin
      run: cat
      use_stdin: true
```

and it works.

PR #737, opened 2024-05-30
perf: delay git and uname commands in hook scripts until needed

scop:
Closes # (issue)

#### :zap: Summary

Avoid running git and uname commands in scripts until their output is actually needed.

I suggest reviewing without showing whitespace changes due to changes in indentation.

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation


mrexox:
Thank you! This makes sense

PR #748, opened 2024-06-14
chore: setting proper error message for missing lefthook file

Cadienvan:
#### :zap: Summary

While using lefthook I found out the error message being shown was incorrect.
Two negations have been put in the same sentence.

I wanted to help, so here you have it :D

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation


mrexox:
Thank you! Good error messages are really important for CLI tool!

Issue #775, opened 2024-07-18
git commit command with error in macos `14.4.1` for M3.


KomachiSion:
### :wrench: Summary

git commit command with error in macos `14.4.1` for M3.

```
╭──────────────────────────────────────╮
│ 🥊 lefthook v1.7.3  hook: pre-commit │
╰──────────────────────────────────────╯
npx: installed 1 in 1.106s
npx: command not found: lefthook-darwin-x64
```

### Lefthook version

```
 npx lefthook version 
1.7.3
```

### Steps to reproduce

run git commit command with lefthook 1.7.3

### Expected results

commit successfully

### Actual results

error: `npx: command not found: lefthook-darwin-x64`

### Possible Solution

force downgrade to 1.7.2.
```
npm install lefthook@1.7.2 --save-dev
```

### Logs / Screenshots

```
19:58:44.675: [nacos] git -c credential.helper= -c core.quotepath=false -c log.showSignature=false checkout HEAD -- pom.xml
19:58:57.931: [nacos] git -c credential.helper= -c core.quotepath=false -c log.showSignature=false add --ignore-errors -A -f -- pom.xml
19:58:57.985: [nacos] git -c credential.helper= -c core.quotepath=false -c log.showSignature=false commit -F /private/var/folders/yj/8c9m_w0s4hv6zvryj59r75lr0000gp/T/git-commit-msg-.txt --
╭──────────────────────────────────────╮
│ 🥊 lefthook v1.7.3  hook: pre-commit │
╰──────────────────────────────────────╯
npx: installed 1 in 1.106s
npx: command not found: lefthook-darwin-x64
```

luizwhite:
similar issues here https://github.com/evilmartians/lefthook/issues/773
I replied that the error still occurs

amochkin:
Try 1.6.19, hope it helps.

KomachiSion:
I think this bug is from PR https://github.com/evilmartians/lefthook/pull/774. 

but https://github.com/evilmartians/lefthook/pull/774 close #733, maybe different.

KomachiSion:
> Try 1.6.19 or 1.6.18, hope it helps.

I force downgrade to 1.7.2, it work well.
Before 1.7.3 released and auto-upgrade, git commit never failed.

mrexox:
Hey! Please checkout 1.7.4 release. This issue must be fixed there.

PR #776, opened 2024-07-19
fix: rollback packaging changes

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/775

**:wrench: Summary**

Looks like optional dependencies (os-arch packages) conflict with the original `lefthook`. This was the only significant change that probably broke the behavior of `npx lefthook`.

### DIFF
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -75,7 +75,7 @@ call_lefthook()
       mint run csjones/lefthook-plugin "$@"
     elif command -v npx >/dev/null 2>&1
     then
-      npx lefthook-${osArch}-${cpuArch} "$@"
+      npx lefthook "$@"
     else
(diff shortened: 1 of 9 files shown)


dalisoft:
@mrexox Yes, it may cause such issue if two or more packages in same node_modules has same bin name. The solution was remove binary from main package and call binary from optional package but only architecture supported one package will be installed, so only that package could be called. 

If there was full guide how to test locally or makefile/Dockerfile to test locally (automatically), i’d test.

no need for npx lefthook-{os}, enough of lefthook as lefthook binary was defined in optional package

mrexox:
Sounds good, so let's try one more time with `"bin": "lefthook"` removed from the original package?

dalisoft:
Yes, that was method to reduce startup time

PR #796, opened 2024-07-29
fix: revert packaging change

mrexox:
**:wrench: Summary**

Revert https://github.com/evilmartians/lefthook/pull/794


mrexox:
@dalisoft, this didn't work, so I am reverting.

dalisoft:
I am working on new approach, which should solve that case. npx or bunx does not work with current direct binary approach

PR #705, opened 2024-04-15
perf(`npm`): startup time reduce

dalisoft:
Closes #703 

#### :zap: Summary

This PR should improve startup performance to get even more out of your performant library

#### :ballot_box_with_check: Checklist

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation

Issue #703, opened 2024-04-13
Performance for `npm` could be even more faster?


dalisoft:
### :zap: Summary

Hi @evilmartians team/contributors

Tuning `npm` package could lead into performance improvements and reducing startup execution.

Opened similar PR/Issue on other projects but impossible to check locally due of hard running local or lacking of running locally easier script.

- https://github.com/dprint/dprint/pull/839/files & https://github.com/dprint/dprint/pull/840/files
- https://github.com/biomejs/biome/pull/2359/files
- https://github.com/oxc-project/oxc/pull/2920/files

As these PR not merged into neither of them, so first decided to ask from you

### Value

| Command                     |   Mean [ms] | Min [ms] | Max [ms] |    Relative |
| :-------------------------- | ----------: | -------: | -------: | ----------: |
| `../lint-staged.sh`         | 154.3 ± 2.8 |    151.1 |    159.3 |        1.00 |
| `lint-staged`               | 306.9 ± 6.1 |    300.3 |    320.8 | 1.99 ± 0.05 |
| `lefthook run pre-commit`   | 202.8 ± 4.4 |    197.3 |    210.1 | 1.31 ± 0.04 |
| `./lefthook run pre-commit` | 162.7 ± 3.2 |    160.4 |    170.7 | 1.05 ± 0.03 |

### Behavior and configuration changes

If `npx` is not important, can be tuned for using `lefthook` binary from `optionalDependencies` or prepare script can download and change binary for main `npm` package `bin` field.

### Test script locally for you

<details>
<summary><code>postinstall.js</code></summary>

```js
import { prepare, maps } from "binary2npm";

maps.os = {
  linux: "Linux",
  freebsd: "Freebsd",
  darwin: "MacOS",
  win32: "Windows",
};
maps.arch = {
  arm64: "arm64",
  x64: "x86_64",
  x86: "x86_64",
};

await prepare({
  remote: "github",
  author: "evilmartians",
  repository: "lefthook",
  remoteToken: process.env.GITHUB_TOKEN,
  binary: "lefthook",
  orders: ["binary", "version", "os", "arch"],
});

```
</details>

- Try run `node postinstall.js`
- Then benchmark direct running versus using custom `bin.js` from `npm` package

mrexox:
[...]

Issue #688, opened 2024-04-01
npm: Not working with bun without node


cethien:
### :wrench: Summary

seems like the tool doesn't work without node. i have only [bun](https://bun.sh) for js on my system. 
after `lefthook install`, every hooked git actions results in `/usr/bin/env: ‘node’: No such file or directory`

### Lefthook version

1.6.7 599459eb87d1dab4fd56707fd1ad3a4209c133c8

### Steps to reproduce

```bash
curl -fsSL https://bun.sh/install | bash
cd <repo>
bun add -D lefthook
<define any lefthook action, eg. commitlint>
lefthook install
git add . && git commit -m "hello"
```

### Expected results

commitlint error

### Actual results

/usr/bin/env: ‘node’: No such file or directory

### Logs / Screenshots

```bash
+ '[' '' = 0 ']'
+ call_lefthook run prepare-commit-msg .git/COMMIT_EDITMSG message
++ git rev-parse --show-toplevel
+ dir=<userprofile>/source/posh-nv
++ uname
++ tr '[:upper:]' '[:lower:]'
+ osArch=mingw64_nt-10.0-22631
++ uname -m
++ sed s/aarch64/arm64/
+ cpuArch=x86_64
+ test -n ''
+ lefthook.exe -h
+ lefthook.bat -h
+ test -f <userprofile>/source/posh-nv/node_modules/lefthook/bin/index.js
+ <userprofile>source/posh-nv/node_modules/lefthook/bin/index.js run prepare-commit-msg .git/COMMIT_EDITMSG message     
/usr/bin/env: ‘node’: No such file or directory
```

dalisoft:
@mrexox @cethien #705 maybe fixes this case as [removes](https://github.com/evilmartians/lefthook/pull/705/files) require of Node.js and `#!/usr/bin/env node`

Blaquewithaq:
For those of you, that want to utilize this package with bun now, Ive figured out a quick patch solution until bun support has been officially added.

Add these to your package.json file and you'll be good to go. 
```
"prepare:lefthook": "lefthook install && bun -e \"const fs=require('fs'); fs.writeFileSync('node_modules/lefthook/bin/index.js', fs.readFileSync('node_modules/lefthook/bin/index.js', 'utf8').replace(/^#!\\/usr\\/bin\\/env\\s+node/gm, '#!/usr/bin/env bun'))\"",
		"postinstall": "bun prepare:lefthook"
```

[...]

PR #798, opened 2024-07-30
chore(`npm`): use `bin.sh` for execute

dalisoft:
Related issues and PR: #688, #703, #705, #776

And #794, #796 

#### :zap: Summary

Reduce execution startup time

This should solve few issues related to bun and it has a `bunx`/`npx` support compared to https://github.com/evilmartians/lefthook/pull/705
**This script, `bin.sh` executes only as later it will be replaced to actual `binary` file**
This approach does not work on `pnpm` and does not reduce startup time.  For `npx` it is faster but not much, for `bunx` it is marginally faster

- [ ] Check locally
- [ ] Add tests
- [ ] Add documentation


dalisoft:
@mrexox I was not able to test locally, but really ready to test if you provide any way to test locally for `bunx`/`npx` or other failing case

packaging/npm/lefthook/bin.sh
mrexox, line 1: I am sure that this script solves the problem with `bun`, but it brings some complicated logic we will have to maintain:

- cache dirs that might change in further releases of bun/npx
- replacements in package.json
- recursive search using `find`
- `sed`, `ln`, `rm -rf` which I usually don't expect in a executable from NPM

I do appreciate your work but I have to say that this solution does not fit my vision of how lefthook should be distributed. 

I like the idea of moving from JS script to shell script, but I think it should be very straightforward even if it is not very fast to execute.

dalisoft: Yeah, you right. It might not work properly later if one of platform changes it's cache or directory

Issue #502, opened 2023-06-16
Provide an Alpine Linux (aports) package


sanmai-NL:
### :zap: Summary

Provide an Alpine Linux (aports) package, so that developers and DevOps professionals can use Lefthook more easily from within a container. See #222 too.

### Value

Alpine Linux is heavily used in the Linux container space.

### Behavior and configuration changes

Follow https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package and submit the package to Alpine Linux. The role of maintainer needs to be fulfilled (but this shouldn't be much work once the package is first done).

mrexox:
Thank you! This is a good idea. I will take a look.

sanmai-NL:
The [Melange](https://github.com/chainguard-dev/melange) tool claims to make this easier than following the guide on the Alpine Linux wiki.

PR #843, opened 2024-10-16
ci: generate apk files

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/502

**:zap: Summary**

In this PR we only generate the package. Future plan is to add lefthook to Alpine registry

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -131,6 +131,7 @@ nfpms:
     builds:
       - no_self_update
     formats:
+      - apk
       - deb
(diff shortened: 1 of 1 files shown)


HugoImaios:
Nice addition :tada:

> Future plan is to add lefthook to Alpine registry
is there an existing issue for that ? (would like to Subscribe to it)

mrexox:
Yep, here it is: https://github.com/evilmartians/lefthook/issues/844

PR #853, opened 2024-10-21
feat: support globs in extends

mrexox:
Closes https://github.com/evilmartians/lefthook/discussions/852

**:zap: Summary**

Support globs in `extends` option.

- [x] Expand the possible globs before extending the main config
- [x] Add tests

Probably need to change checksum check from the file checksum to a memory object checksum – https://github.com/evilmartians/lefthook/pull/854

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,6 +138,8 @@ no_tty: true
 
 You can extend your config with another one YAML file. Its content will be merged. Extends for `lefthook.yml`, `lefthook-local.yml`, and [`remote`](#remote) configs are handled separately, so you can have different extends in these files.
 
+You can use asterisk to make a glob.
+
 **Example**
 
 ```yml
@@ -148,6 +150,7 @@ extends:
   - /home/user/work/lefthook-extend-2.yml
   - lefthook-extends/file.yml
   - ../extend.yml
+  - projects/*/specific-lefthook-config.yml
 ```
 
 > [!IMPORTANT]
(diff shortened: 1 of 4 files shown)


VanTanev:
Hey, for some reason I didn't get a nofication about your response in the discussion, sorry for getting back to you late!

This is a good step forward, but I think it needs one more improvement to be viable.

Configs that are extended should automatically set the `commdands.{}.root` parameter to the folder of the extended config file.

Example, given:

```yaml
# ./lefthook.yml
extends:
 - "apps/*/lefthook.yml"
```

and
```yaml
# ./apps/store/lefthook.yml
pre-commit:
  commands:
    jest:
      run: pnpm jest --findRelatedTests --passWithNoTests {staged_files}
```

It will fail with:
```
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "jest" not found
```

To fix it, one needs to manually add `root: apps/store`
```yaml
# ./apps/store/lefthook.yml
pre-commit:
  commands:
    jest:
      root: apps/store # <!------ THIS HERE!!!
      run: pnpm jest --findRelatedTests --passWithNoTests {staged_files}
```

Which is very clunky.

mrexox:
Oh, I see. Yes, it might be too much to add root to every command, but now this is a must. I don't think it's a good idea to automatically add `root` because you can merge lots of files from different folders, and this can become too complicated to understand if something is added automatically.

Currently I don't have an elegant solution yet, so I have nothing to suggest except for explicit configuration for every command.

VanTanev:
Yeah, the correct solution is not obvious. For example, `lint-staged` has the reverse approach, where leaf config files import from a root file, so it makes sense that commands are relative to leaf config file locations.

I do think that in the general case, automatically setting `root` for leafs should be fine, but there are definitely edge-cases, and such a change would be semver-breaking.

So, ultimately we're stuck with some repetition.

PR #857, opened 2024-10-25
feat: add refetch_frequency parameter to settings

gabriel-ss:
#### :zap: Summary

First of all, thanks everyone for the great tool!

The `remote` feature of lefthook allows us at the place where I work to easily share configuration across our hundreds of repositories. Usually, our devs have some dozens of repositories cloned on their machines and work actively in one or two at a time. To ensure that they will have a reasonably recent config when going back to a repo after some time, we configure our remotes with `refetch: true`, but this has the side effect that every time they run lefthook on a currently active repository, either by commiting something or by hand to manually check the output of the QA tools, they trigger a fetch.

To make the experience smother, this PR introduces a new parameter `refetch_frequency` that allows a more fine grained control over the remote refetching by allowing the definition of time periods instead of just a boolean flag. This is a working implementation already, but it defers some decisions that could be made to maybe improve the usability. I didn't want to touch the following points without consulting the core team before, but:
- I think that this parameter superseds the current `refetch`, so maybe it could be marked deprecated;
- I used ParseDuration from go's stdlib in the current implementation, but a more user-friendly format could be used at the cost of a new dependency.

I also apologize in advance for any oddities in the code, I'm not a go programmer, but I can make changes that you see fit to make the PR more fitting.

- [x] Check locally
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -33,6 +33,7 @@ Config options:
   - [`git_url`](#git_url)
   - [`ref`](#ref-1)
   - [`refetch`](#refetch)
+  - [`refetch_frequency`](#refetch_frequency)
   - [`configs`](#configs)
 - [`<pre-commit>`](#hook-name) hook name
   - [`files` (global)](#files-global)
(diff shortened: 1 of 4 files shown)


mrexox:
Wow! Thank you for providing this PR! I will review it this week and will release if everything is good :+1:

mrexox:
Check out 1.8.2 version!

Issue #132, opened 2020-06-16
Order of scripts and command


lionskape:
Are there any way to control the sequence of scripts and commands?

For example to execute scripts after commands?
Or to execute the first command, after that execute the first script, and after that execute the second command?

Arkweid:
Only with this way: [piped-option](https://github.com/Arkweid/lefthook/blob/master/docs/full_guide.md#piped-option)
Userful for migrations and similar stuff.

lionskape:
Is it possible to control order between commands and scripts?
Like - command_1, script_1, command_2

Piped example shows only order between commands.

Arkweid:
No, only commands.

sanmai-NL:
New location of docs: https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#piped

BnGx:
Currently the hooks are executed in alphabetical order, however one of the following options would instead be appropriate:

- Leave the definition order unchanged (including `interactive` hooks and those extended with the `extends` option)
- Insert a specific option (`priority` or `order`) that has a default value (for example 50) and that it can be changed in order to modify the order of execution.

```yaml
database:
  piped: true
  commands:
    create:
      priority: 40
      run: rake db:create
    migrate:
      run: rake db:migrate
    seed:
      priority: 60
      run: rake db:seed
```

This would improve the understanding of the lefthook configuration without introducing distortions.

mrexox:
@BnGx , interactive hooks always run after non-interactive hooks. Scripts run after the all commands. The order is alphanumeric, so you can do something like

```yml
hook:
  commands:
    1_do_this:
      run: ...
    2_then_this:
      run: ...
    ...
```

It probably makes sense to allow a special order of scripts + commands, but why? Would you describe a case when this is really needed and alphanumeric order is not enough?

BnGx:
As I see it, implementing a priority option in Lefthook would be a better approach for several reasons:

- **Enhanced Flexibility and Precision**: A priority-based system would enable developers to accurately determine the execution sequence of hooks, which is vital in complex projects where the order can substantially affect both quality and workflow efficiency.
- **Greater Scalability**: The alphabetical method tends to become unwieldy and disorganized with an increasing number of hooks. Conversely, a priority-based approach would remain clear and manageable, even in larger projects.
[...]

PR #861, opened 2024-10-31
feat: add jobs option

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/132

**:zap: Summary**

- [x] Add `jobs` and `group` new options, so recursive piped/parallel flows could be built
- [x] Implement the same run logic for `jobs` as for commands and scripts
- [x] Properly merge `jobs` options with `lefthook-local.yml` and extends
- [x] Add unit tests
- [x] Add integration tests
- [x] Add docs

**Old syntax**

```yaml
pre-commit:
  parallel: true
  commands:
    lint:
      run: yarn lint
      stage_fixed: true
    fix-typos:
      run: typos
    migrate:
      run: lefthook run migrate
  scripts:
    "graphql-generate.sh":
      runner: bash

migrate:
  piped: true
  commands:
    setup:
      run: yarn setup
      priority: 1
    migrate:
      run: yarn migrate
      priority: 2
```

**New syntax**

```yaml
pre-commit:
  parallel: true
  jobs:
    - name: lint
      run: yarn lint
      stage_fixed: true
    - run: typos
    - script: graphql-generate.sh
      runner: bash
    - name: migrate 
      group:
        piped: true
        jobs:
          - run: yarn setup
          - run: yarn migrate
```


VanTanev:
Hey, a question here: Re `do[].group.do[]` items the same as `do[]` items?

In this case, the nested items don't have a name - is it valid for the top level items to also omit a name, and what happens then, is the name just the value of `run`?

mrexox:
Yes, `name` is optional. The only mandatory options are either `run`, `script`, or `group`. `run` or `script` value will be used if `name` is empty, and for a group without a name it will be the index of a group - `[3]`

PR #879, opened 2024-11-26
Option to fully disable LFS hooks

zachahn:
https://github.com/evilmartians/lefthook/discussions/821

#### :zap: Summary

Git LFS is an opt-in feature that lets people store large files outside a git repository, saving only pointers to them in the git repository itself.

Git does not use LFS by default. One must manually specify files and file patterns using the `.gitattributes` file.

I've found that LFS adds about 2.2 seconds when running lefthook in my employer's large git repo.

<details>
<summary>My small benchmark</summary>

```
❯ /usr/bin/time git lfs post-checkout e6f8d32b e6f8d32b 0
        2.24 real         0.39 user         1.51 sys
❯ /usr/bin/time git lfs post-checkout e6f8d32b e6f8d32b 0
        2.22 real         0.39 user         1.50 sys
❯ /usr/bin/time git lfs post-checkout e6f8d32b e6f8d32b 0
        2.28 real         0.41 user         1.51 sys
❯ /usr/bin/time git lfs post-checkout e6f8d32b e6f8d32b 0
        2.29 real         0.41 user         1.51 sys
❯ /usr/bin/time git lfs post-checkout e6f8d32b e6f8d32b 0
        2.30 real         0.42 user         1.53 sys
```

</details>

Considering the ~500 engineers in my company, and assuming an average of 1 `git checkout` per day (I assume this is a low estimate, but I don't have metrics on this), _**we'd save 6.1 hours of engineering time every month.**_

##### NOTE

I'd personally recommend that we rework this feature—that we change the flag to `enable_lfs` and disable LFS hooks by default. I'm more than happy to do so.

Because LFS is fully opt-in, and because of its poor performance in large repositories, I don't believe it makes sense to enable this feature by default.

##### ALSO NOTE

I disabled LFS for this project as well since it's not needed here (see "chore" commit). I assume you will want me to revert that change! I'm more than happy to 😅.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/.lefthook.toml
+++ b/.lefthook.toml
@@ -1,3 +1,5 @@
+skip_lfs = true
+
 [pre-commit]
 parallel = true
 
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -70,6 +70,11 @@ func (run) New(opts *lefthook.Options) *cobra.Command {
 	)
 
 	runCmd.Flags().BoolVar(
+		&runArgs.SkipLFS, "skip-lfs", false,
+		"skip running git lfs",
+	)
+
+	runCmd.Flags().BoolVar(
 		&runArgs.FilesFromStdin, "files-from-stdin", false,
 		"get files from standard input, null-separated",
 	)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -37,6 +37,7 @@ type Config struct {
 	NoTTY                   bool        `mapstructure:"no_tty,omitempty"`
 	AssertLefthookInstalled bool        `mapstructure:"assert_lefthook_installed,omitempty"`
 	Colors                  interface{} `mapstructure:"colors,omitempty"`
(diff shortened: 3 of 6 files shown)


mrexox (commented):
Hey! Thank you for this PR! I like this change, but I have a concern on naming. Could you call it "skip lfs" everywhere (config, arg, run option)?

zachahn:
Done! I added a simple test too, but it doesn't really test anything besides going through the new code path. Let me know if you'd like any other changes.

Just to confirm, I'm guessing you might not want to flip the default and disable LFS unless explicitly enabled? I would love to bring this improvement to all repositories that use lefthook, but I understand that could be classified as a breaking change. But of course, I'm happy with this PR as is :)

Thanks for reviewing!

mrexox (approved):

Issue #846, opened 2024-10-16
Make `parallel: true` work with priorities


scop:
### :zap: Summary

Would good to have `parallel: true` work with priorities.

### Value

Faster runs containing priorities (~ dependencies).

### Behavior and configuration changes

For example:
```yaml
foo:
  parallel: true
  commands:
    cmd-a:
      priority: 1
    cmd-b:
      priority: 2
    cmd-c:
      priority: 3
    cmd-d:
      # no priority set
    cmd-e:
      priority: 1
    cmd-f:
      priority: 3
```
...and `lefthook run foo` would result in the commands to be run in parallel _within groups determined by priority_.

So the order of execution would go in sequence
1. priority 1: cmd-a and cmd-e in parallel
2. priority 2 (after prio 1 jobs complete): cmd-b alone
3. priority 3 (after prio 2 jobs complete): cmd-c and cmd-f in parallel
4. priority +Infinity (after prio 3 jobs complete): cmd-d alone

scop:
That looks otherwise nice and useful, but I suppose with the jobs config above, it would no longer be possible to run specific commands only?

With `commands` one could do `lefthook run pre-commit --commands COMMANDNAME` but there's no corresponding option to run only some job or group names?

mrexox:
Hmm, nice note. I think it should be possible to add the same `--jobs` argument. But if you have jobs with the same name – all of them will be run. Do you find this useful? I think I can add it in future releases.

mrexox:
I guess this issue can be closed since now lefthook supports `jobs` and `group` where you can use any level of piped/parallel execution.

PR #904, opened 2024-12-21
feat: add ability to specify job names to run command

mrexox:
Idea from discussion of https://github.com/evilmartians/lefthook/issues/846

**:zap: Summary**

- [x] Add `lefthook run <hook> --jobs job1 job2`
- [x] Add tests

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -94,6 +94,11 @@ func (run) New(opts *lefthook.Options) *cobra.Command {
 		"run only specified commands",
 	)
 
+	runCmd.Flags().StringSliceVar(
+		&runArgs.RunOnlyJobs, "jobs", nil,
+		"run only specified jobs",
+	)
+
 	err := runCmd.Flags().MarkDeprecated("files", "use --file flag instead")
 	if err != nil {
 		log.Warn("Unexpected error:", err)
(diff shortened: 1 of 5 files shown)


scop:
From https://github.com/evilmartians/lefthook/issues/846#issuecomment-2558039200_
> But if you have jobs with the same name – all of them will be run. Do you find this useful?

Yeah, I think that would be fine. And also as mentioned in #846, it would be great if one could also run (only) groups by their name.

I'm also wondering if all those three (commands, jobs, and groups) could be merged into one option, let's say `--only THING` would run only all commands, jobs, and/or groups named `THING`. Not sure for example if commands and jobs can coexist, or if there's much use for three separate options for these as they're kind of mutually exclusive anyway (unless e.g. `--commands FOO --jobs BAR` would run commands `FOO` _and_ jobs `BAR`. Anyway mostly jut thinking aloud, not refined ideas in any way.

mrexox:
I'd keep both `--commands` and `--jobs`. If you want to avoid running commands or jobs, you can pass `--commands no-command-with-this-name` or `--jobs no-job-with-this-name`.

I think I will try deprecating commands in the future, but I don't want to break the compatibility too early.

PR #913, opened 2025-01-08
feat: run --jobs completion

scop:
#### :zap: Summary

Adds `lefthook run foo --jobs <TAB>` completion, like there currently exists for `--commands`.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation


mrexox:
Thank you! This won't reach the nested jobs but it's OK.

Issue #912, opened 2025-01-07
Inherit `exclude` from parent jobs


scop:
### :zap: Summary

It would seem that `glob` is inherited by jobs from their parent jobs, which is nice. However, exclude does not seem to work that way.

### Value

Less boilerplate and repeated config.

For example, let's say config that formats and check shell scripts:
```yaml
pre-commit:
  jobs:
    - name: linters
      group:
        parallel: true
        jobs:
          # bunch of other jobs go here
          - name: shell-tools
            glob: "*.{bash,sh}"
            group:
              jobs:
                - name: shfmt
                  exclude:
                    - vendor/*
                  run: shfmt --write --simplify {staged_files}
                - name: shellcheck
                  exclude:
                    - vendor/*
                  run: shellcheck --format=gcc {staged_files}
```

The `glob` does seem to apply from the `shell-tools` level down to `shfmt` and `shellcheck`, but the `exclude`s need to be duplicated in both to work.

If/when implemented, this would also at least alleviate if not solve https://github.com/evilmartians/lefthook/issues/593 as one could have a top level job with the desired excludes in it.

### Behavior and configuration changes

Make jobs' `exclude` inherit entries from parent level jobs. Child jobs' `exclude` entries should be added to (not replaced with) the ones inherited from parent jobs.

PR #916, opened 2025-01-09
feat: inherit exclude option in groups

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/912

**:zap: Summary**

- [x] Inherit `exclude` option: append new values if array given, overwrite with the new regexp otherwise
- [x] Add a test
- [x] Update the docs

### DIFF
--- a/docs/mdbook/configuration/jobs.md
+++ b/docs/mdbook/configuration/jobs.md
@@ -25,7 +25,7 @@ pre-commit:
 **Job Groups**
 
 - Groups can include other jobs.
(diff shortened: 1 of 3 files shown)


scop:
Thanks for this! It seems to work, but only one level deep, i.e. inherits only from the immediate parent group, not from  grandparent groups. I wonder if that's on purpose, or on the TODO, or...?

mrexox:
Oh, indeed. That's a bug, I'll fix it in next release :+1:

PR #941, opened 2025-01-30
Document the unusual behaviour of ** in glob option

jamiecobbett:
#### :zap: Summary

I had assumed that glob patterns work like they do with the *nix `ls` command, or in tools like lint-staged. I can imagine people expecting it to behave the same.

It seems that `**` is interpreted as a single asterisk, ie meaning "1+ directories deep", rather than "0 or more directories deep".

Whilst there is a link to the Go glob library used, and that README mentions double asterisks, I was mainly left confused by it:

> Syntax is inspired by [standard wildcards](http://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm), except that ** is aka super-asterisk, that do not sensitive for separators.

I don't think this covers the difference I'm bringing up. I also don't know what is meant by "super-asterisk" here - I'd love to know!

More details, including an example are in the commit message.

Thanks for making such an awesome tool!

### DIFF
--- a/docs/mdbook/configuration/glob.md
+++ b/docs/mdbook/configuration/glob.md
@@ -31,6 +31,10 @@ pre-commit:
 
 For patterns that you can use see [this](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm) reference. We use [glob](https://github.com/gobwas/glob) library.
 
+***When using `root:`***
+
+Globs are still calculated from the actual root of the git repo, `root` is ignored.
+
 ***Behaviour of `**`***
 
 Note that the behaviour of `**` is different from typical glob implementations, like `ls` or tools like `lint-staged` in that a double-asterisk matches 1+ directories deep, not zero or more directories.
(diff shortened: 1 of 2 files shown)


jamiecobbett:
I've also noticed that if you use the `root:` option, then it seems the paths in `glob:` don't take that into account - ie you must specify the full path, whereas I assumed it would be relative to the `root`.

If that's correct, I'm happy to add that to this PR, or open a separate one.

docs/mdbook/configuration/glob.md
```diff
@@ -31,6 +31,23 @@ pre-commit:
...
+If you want to match *both* files at the top level and nested, then rather than:
+
+```yaml
+glob: "src/**/*.js"
+```
+
+You'll need something like:
+
+```yaml
+glob: "{src/*.js,src/**/*.js}"
```
jamiecobbett, line 46: On further testing, I'm fairly sure you just need `glob: "src/*.js"`

jamiecobbett: OK, updated

jamiecobbett:
> I've also noticed that if you use the root: option, then it seems the paths in glob: don't take that into account - ie you must specify the full path, whereas I assumed it would be relative to the root.

I've gone ahead and added that to this PR, but please shout if you'd prefer it to be broken off.

PR #942, opened 2025-01-30
fix(jsonschema): add $schema property

mst-mkt:
#### :zap: Summary

In the current jsonschema, specifying `$schema` results in a warning as shown in the following image.

This is because `"additionalProperties": false` and there is no `$schema` in `properties`.

image

To resolve this, the `$schema` property was added.  
This approach has been observed in other projects as well, such as [@biomejs/biome](https://github.com/biomejs/biome/blob/738ad51d654af10d5dd4faf5e8f8bdef5db81f09/packages/%40biomejs/biome/configuration_schema.json#L7-L10).

```diff
    "properties": {
+     "$schema": {
+       "type": "string"
+     },
    },
```

I am not very familiar with Go, so I would appreciate any corrections or suggestions.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/gen/jsonschema.go
+++ b/internal/gen/jsonschema.go
@@ -18,6 +18,7 @@ func main() {
 	r.AdditionalFields = func(t reflect.Type) []reflect.StructField {
 		if t == reflect.TypeOf(config.Config{}) {
 			return reflect.VisibleFields(reflect.TypeOf(struct {
+				Schema               string      ` json:"$schema,omitempty"`
 				PreCommit            *config.Hook `json:"pre-commit,omitempty"`
 				ApplypatchMsg        *config.Hook `json:"applypatch-msg,omitempty"`
 				PreApplypatch        *config.Hook `json:"pre-applypatch,omitempty"`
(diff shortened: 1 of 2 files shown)


mrexox:
Hey! Could you help me understand how does it biome use the `$schema` field? And are you adding this field to `lefthook.json`? Currently lefthook does not handle `$schema` parameter internally

mst-mkt:
@mrexox Thank you for your comment

> Could you help me understand how does Biome use the $schema field?

The `$schema` field is not specifically used by Biome itself. It's a standard JSON Schema keyword that enables editor support in tools like VSCode, providing features such as auto-completion and validation when editing configuration files. I mentioned Biome only as an example of a project that includes `$schema` in its schema definition.

> And are you adding this field to lefthook.json? Currently lefthook does not handle $schema parameter internally

Yes, I'm adding the `$schema` property to improve editor support, not for Lefthook's internal use. While Lefthook itself doesn't process this field, having it in the schema enables better tooling support in editors like VSCode and JetBrains IDEs.

information about specifying `$schema`, you can refer to:
- [VSCode JSON Language Server Docs](https://github.com/Microsoft/vscode/blob/main/extensions/json-language-features/server/README.md#schema-configuration-and-custom-schema-content-delivery)
- [JetBrains JSON Schema Guide](https://www.jetbrains.com/guide/dotnet/tips/json-schema/)

PR #956, opened 2025-02-19
Speed up

judofyr:
#### :zap: Summary

Some smaller speed-ups. In one random example:

* With no edited/staged files it went from **70 ms** to **42 ms**.
* With a single staged file it went from **155 ms** to **90 ms**.
* When running on all files it went from **200 ms** to **180 ms**.

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation


judofyr:
Ah, I see that tests are failing since they’re not using the Repository constructor. I can fix it!

judofyr:
Okay, _now_ I think we're good. I wasn't aware of the integrity tests. Those are actually very nice! Changes since last time:

- Fixes a bug `stagedFilesWithDeletedOnce`. It was calling `FindExistingFiles` instead of `FindAllFiles`. This was caught by an integrity test.
- Change the pre-compute to also wait for when it's done. This is needed just so that we don't hold any locks on the Git repo.
- Only do pre-computation in `lefthook run`.

internal/git/repository.go
```diff
@@ -94,38 +98,73 @@ func NewRepository(fs afero.Fs, git *CommandExecutor) (*Repository, error) {
...
+	})
+
+	eg.Go(func() error {
+		_, _ = r.statusShortOnce()
+		return nil
+	})
+
+	return func() {
+		_ = eg.Wait()
+	}
```
mrexox, line 141: I think a simple WaitGroup would fit here better since we do not track errors

internal/git/repository.go
```diff
@@ -94,38 +98,73 @@ func NewRepository(fs afero.Fs, git *CommandExecutor) (*Repository, error) {
...
+}
+
+// InitializeForTest must be called if you constructed a Repository without using `NewRepository`.
+// This can also be called multiple times to reset the cache.
+func (r *Repository) InitializeForTest() {
```
mrexox, line 146: Could you call this function differently, please? We use it in the New func, so it's not only for test. It can be `SetupStageCommands` or `SetupCachedCommands`, or even a mysterious `Setup` :)

internal/git/repository.go
judofyr: Fair enough!

internal/git/repository.go
judofyr: I opted for just `Setup`. In hindsight I realize that `SetupCache` is maybe more precise, but since it's all in `internal/` I'm not sure it matters.

judofyr:
I addressed your comments in the latest commit and also ended up fixing linting errors directly in some earlier commits. Both tests and lints _should_ now pass in all the four commits.

I can squash the last commit if you want.

mrexox (approved):

mrexox:
Thank you for preparing this PR!

judofyr:
Thanks for quick reviews and a well-maintained project!

judofyr:
Oh, you might also find this interesting: https://t.holmium.no/dia/hk-vs-lefthook/. This PR was triggered by me looking more into the performance claims of [hk](https://hk.jdx.dev/about.html). This article contains a flamegraph of Lefthook execution and some more timings.

mrexox:
Thank you for sharing this article! This is really cool!

PR #965, opened 2025-03-01
feat: Add shell options to run commands

### LINKED ISSUE #953: Wrap commands in Git bash on Windows
### :zap: Summary

According to [docs](https://lefthook.dev/configuration/run.html?highlight=windows#run), `run` commands behave differently on Linux and Windows. This is not ideal.

We can wrap commands in git bash, which is located in `C:\Program Files\Git\bin\sh.exe`.

### Value

Provide command consistency.

commands should behave the same on Linux and Windows.

NavyD:
Closes #953

#### :zap: Summary

Add the `shell []string` option to the jobs/commands hook to run the command to provide consistency for Windows and Unix, and the default is empty will not cause compatibility issues.

```yaml
pre-commit:
  jobs:
    - name: test
      shell:
        - powershell
        - -Command
      run: |
        echo test
        $PSVersionTable
```

Note: This is just a simple implementation, any suggestions for improvement are welcome, I'm not very familiar with golang, so I hope this doesn't bother you!

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/config/command.go
+++ b/internal/config/command.go
@@ -7,6 +7,8 @@ import (
 var ErrFilesIncompatible = errors.New("One of your runners contains incompatible file types")
 
 type Command struct {
+	Shell []string `json:"shell,omitempty"    jsonschema:"oneof_type=string;array" mapstructure:"shell"      toml:"shell,omitempty"    yaml:",omitempty"`
+
 	Run   string `json:"run"             mapstructure:"run"   toml:"run"             yaml:"run"`
 	Files string `json:"files,omitempty" mapstructure:"files" toml:"files,omitempty" yaml:",omitempty"`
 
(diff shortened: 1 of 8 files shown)


mrexox:
Hey! Thank you for creating this PR. I pretty like this option but I could resolve the issue without adding the additional configuration options.

I think to take a pause and revisit this idea, maybe when I found a new issue this change can fix I will finish this PR.

NavyD:
Thank you for your reply!

I created this PR for a better windows experience, to be able to run some simple commands directly instead of creating script files to execute them! Fixing this issue is just by the way, you can ignore or close this PR!

mrexox:
Closing the PR since found another approach and closed the original issue

Issue #989, opened 2025-04-03
Version 1.11.5 introduces exception for lefthook CLI when git bash doesn’t exist in C:Program Files\Git\bin\sh.exe


davidensinger:
### :wrench: Summary

The Windows fix #979 in version 1.11.5 introduces an error for Windows dev environments that have git bash installed in a custom location when lefthook is run via CLI. This causes git hooks not to run. The functionality when run within the actual git hooks works as expected. I didn’t actually test version 1.11.5, but I did test version 1.11.4 and version 1.11.6.

### Lefthook version

1.11.6

### Steps to reproduce

With version 1.11.6:

```shell
lefthook run pre-commit --force --verbose --no-tty
```
Results in this error:

```shell
fork/exec C:Program Files\Git\bin\sh.exe: The system cannot find the path specified
```

I also see this log:

```shell
[lefthook] run: "C:\Program Files\Git\bin\sh.exe" -c "prettier --write  && git add "
``` 

### Expected results

No error message and Prettier runs.

### Actual results

See reproduction steps above.

### Possible Solution

Downgrade to 1.11.4:

```shell
lefthook run pre-commit --force --verbose --no-tty
```

Results in this log and no error:

```shell
[lefthook] executing: prettier --write  && git add
Nothing specified, nothing added.
...more log output...
summary: (done in 10.37 seconds)
✔️prettier
``` 

Is there a way to get the installed location of git bash so that instead of hardcoded `C:\Program Files\Git\bin\sh.exe` we get the correct path? I am not knowledgeable about whether there‘s a way to get that info, unfortunately.

### Logs / Screenshots

See above.

Thank you for this wonderful tool, btw.

mrexox:
Hey! Thank you for creating this issue. Could you share the path you have Git installed to? I was afraid that this could break manual execution of `lefthook run` on Windows but I assumed that `C:\Program Files\Git` was widely used. I think I could try adding a fallback or maybe an option to overwrite this value in the config. Since I'm not actively using Windows, I'd appreciate your ideas on this problem.

davidensinger:
So my developer environment is atypical. It’s a virtual Windows environment and git is installed in a different directory than what is convention. I think it’s truly custom where it is installed.

Essentially:

`C:\some-dir\another-dir\yet-another-dir\git\bin\sh.exe`

So that’s not very helpful, I know. I am also not a primary Windows developer.

[...]

PR #992, opened 2025-04-05
fix: fix command execution error on Windows #989

atsushifx:
Issue # (989)

#### :zap: Summary

- changed:   internal/lefthook/runner/exec/execute_windows.go
  - fix: fullPathSh: change `C:\Program Files\git\bin\sh.exe` to `sh.exe` for git bash installed in other directories
  - fix: execute: cmdStr is dont escpape filename double quote, so replace these to escaped double quote

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/lefthook/runner/exec/execute_windows.go
+++ b/internal/lefthook/runner/exec/execute_windows.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	"github.com/evilmartians/lefthook/internal/log"
@@ -16,7 +17,7 @@ import (
 )
 
 const plainSh = "sh"
-const fullPathSh = `C:\Program Files\Git\bin\sh.exe`
+const fullPathSh = `sh.exe`
 
 type CommandExecutor struct{}
 type executeArgs struct {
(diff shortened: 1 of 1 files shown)


mrexox:
Hey! Thank you for making this PR. But `sh.exe` is not available without changes to PATH variable. By default Git installation on Windows only adds the git folder with `git.exe`, `git-lfs.exe`, etc. and `sh.exe` is not there.

I'm afraid there is no way to fix it easily 😕

atsushifx:
Add two functions below to search sh.exe

- `getFullPathSh`
- `findExecutableDir`

These functions attempt to locate `sh.exe` via commonly installed Git locations:

#### Fallback search logic
1. If `git-bash.exe` is found, `sh.exe` is assumed to be in its `bin` subdirectory:
   ```
   .../git-bash.exe → .../bin/sh.exe
   ```

2. If `git.exe` is found under the `cmd` directory (e.g., `.../cmd/git.exe`), 
   `sh.exe` is assumed to be in the sibling `bin` directory:
   ```
   .../cmd/git.exe → .../bin/sh.exe
   ```

3. As a final fallback, the default Git install path is checked:
   ```
   C:\Program Files\Git\bin\sh.exe
   ```

This approach increases compatibility, especially in Windows environments where `sh.exe` is bundled with Git but not exposed globally in `PATH`.

---

Let me know if you'd like to see a test case or additional logging for this logic. 😊

mrexox:
Thank you! That looks much better and I like your idea! I will merge it and try to slightly improve the performance by using `sync.OnceValue`

PR #1001, opened 2025-04-09
feat(npm): allow installing hooks in CI

caugner:
#### :zap: Summary

Allows installing hooks in CI (if it sets `CI=true` automatically) by setting `LEFTHOOK=1` or `LEFTHOOK=true` (recommended), or overriding it with `CI=0` or `CI=false`.

_Background_: MDN recently migrated most of its repositories from husky+lint-staged to lefthook, but we noticed that lefthook behaves different in CI (GitHub workflows): Husky installs hooks by default (which can be disabled with `HUSKY=0`), whereas lefthook doesn't install hooks (if `CI=true` is set). The only workarounds seem to have been setting `CI=""` (which may have side-effects on linting tools), or by running the hooks manually (e.g. via `npx lefthook run pre-commit`).

- [x] Check locally
- [ ] Add tests
- [x] Add documentation

### DIFF
--- a/docs/mdbook/usage/env.md
+++ b/docs/mdbook/usage/env.md
@@ -70,3 +70,8 @@ Set `NO_COLOR=true` to disable colored output in lefthook and all subcommands th
 
 Set `CLICOLOR_FORCE=true` to force colored output in lefthook and all subcommands.
 
+### `CI`
+
+Use `CI=true` to prevent lefthook from installing hooks during the postinstall script.
+
+To override this behavior and allow hook installation despite `CI=true`, set `LEFTHOOK=1` or `LEFTHOOK=true`. This takes precedence over the `CI` variable.
--- a/docs/mdbook/usage/env.md
+++ b/docs/mdbook/usage/env.md
@@ -12,6 +12,16 @@ Use `LEFTHOOK=0 git ...` or `LEFTHOOK=false git ...` to disable lefthook when ru
 LEFTHOOK=0 git commit -am "Lefthook skipped"
 ```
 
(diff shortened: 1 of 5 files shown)


docs/mdbook/usage/env.md
```diff
@@ -70,3 +70,8 @@ Set `NO_COLOR=true` to disable colored output in lefthook and all subcommands th
 
 Set `CLICOLOR_FORCE=true` to force colored output in lefthook and all subcommands.
 
+### `CI`
+
+Use `CI=true` to prevent lefthook from installing hooks during the postinstall script.
```
mrexox, line 75: Could you please add that this is true only for NPM packages?

mrexox: 1. It's OK to use `LEFTHOOK` var
2. I think it makes sense. Currently only presence is being checked and in case of `CI=0` it will pass. We should skip if `CI` value is `0` or `false`.
3. I think it's OK to just check `process.env.CI != '0' && process.env.CI != 'false'`

caugner: Thanks!

I went ahead and added a local `isEnabled` helper (similar to [`isEnvEnabled()` in lefthook.go](https://github.com/evilmartians/lefthook/blob/573fd6f7341fbdabeaddfbf88edbb31ac5441448/internal/lefthook/lefthook.go#L146-L153)) for clarity/readability, but I can inline if you deem it unnecessary.

mrexox (approved):

docs/mdbook/usage/env.md
mrexox: Nice! Thank you!

PR #1004, opened 2025-04-11
add support for uv

toshok:
#### :zap: Summary

Document and support installing `lefthook` in a venv managed by `uv`.

Given the venv isn't activated, `lefthook` won't be in PATH, and we need to check for it the same way as with other package managers.

- [x] Check locally
- [ ] Add tests
- [x] Add documentation


mrexox:
Thank you for creating this PR. I believe that you've tested this change, so I'll just merge it :)

PR #1059, opened 2025-07-02
feat(install): install a specific hook

lucascolley:
Closes #1058

#### :ballot_box_with_check: Checklist

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

```console
lefthook on 🎋 master [🔨?] via  v1.24.4 via 🧚 v0.49.0 (default) 
❯ ls .git/hooks    
ls: .git/hooks: No such file or directory

❯ go run main.go install pre-push
sync hooks: ✔️ (pre-push)

❯ ls .git/hooks                  
pre-push		prepare-commit-msg

❯ go run main.go install foobar
sync hooks: ❌
│  Error: hook not defined: foobar
exit status 1

❯ go run main.go install pre-commit
sync hooks: ✔️ (pre-commit)

❯ ls .git/hooks                    
pre-commit		pre-push		prepare-commit-msg

❯ rm -rf .git/hooks

❯ go run main.go install           
sync hooks: ✔️ (pre-commit, pre-push)

❯ ls .git/hooks     
pre-commit		pre-push		prepare-commit-msg
```

Issue #1058, opened 2025-07-02
feat: `lefthook install {specific-hook}`


lucascolley:
### :zap: Summary

Say my `lefthook.yml` defines two hooks (pre-commit and pre-push). I want to give my team the option to install both (`lefthook install`) or only install one (e.g. `lefthook install pre-commit`)

### Value

Some team members may want to install the pre-commit hook but not the pre-push hook, or vice versa

### Behavior and configuration changes

Is this possible already somehow?

mrexox:
The problem is that when you update your pre-commit hook the lefthook will check the file checksum and resync the whole config, installing all hooks.

The only approach we have now to skip/update some hooks is to use `lefthook-local.yml`, for example

```yml
# lefthook.yml

pre-commit:
  jobs:
    - …

pre-push:
  jobs:
    - …
```

```yml
# lefthook-local.yml

pre-push:
  skip: true
```

In this configuration `pre-push` hook will be skipped locally if someone doesn’t want to use it. However when someone updates `pre-commit` it will be resynced with no issues automatically, but `pre-push` will stay skipped.

Does it solve your case?

lucascolley:
Maybe. Our use-case is https://github.com/prefix-dev/pixi/blob/main/pixi.toml#L118-L119:

```toml
pre-commit-install = "pre-commit install --install-hooks -t=pre-commit -t=pre-push"
pre-commit-install-minimal = "pre-commit install -t=pre-commit"
```

I guess with your suggestion it would be something like:

```toml
pre-commit-install = "lefthook install"
pre-commit-install-minimal = "cp misc/lefthook-local.yml lefthook-local.yml && lefthook install"
```

I guess that works... I can try it out

mrexox:
Cool! Please, don’t forget to put `lefthook-local.yml` into `.gitignore`

lucascolley:
Hmm, it is still quite annoying to have this running on every push, for those who don't want it:

```console
❯ git push                 
✨ Pixi task (lefthook in lint): lefthook run pre-push origin https://github.com/lucascolley/pixi.git: (Run lefthook)
╭──────────────────────────────────────╮
│ 🥊 lefthook v1.11.14  hook: pre-push │
╰──────────────────────────────────────╯
│  pre-push (skip) hook setting
summary: (skip) empty
```

I don't think that is a good solution. We really want customisability of which hooks are installed. Installing them just to skip them every time is suboptimal.

lucascolley:
[...]

PR #1069, opened 2025-07-08
feat: allow installing only specific hooks

mrexox:
Closes https://github.com/evilmartians/lefthook/pull/1059
Closes https://github.com/evilmartians/lefthook/issues/1058

**:zap: Summary**

`lefthook install pre-commit` installs and then syncs only the specified `pre-commit` hook.

**Note**: `lefthook install` will reinstall all hooks


lucascolley:
thanks @mrexox ! Please let me know if I can help with testing. Looking forward to getting rid of pre-commit (the package) 😉

mrexox:
@lucascolley I will release a new version and you can try it then. I think it will be the best way to understand if the feature works good or requires some additional work.

PR #1072, opened 2025-07-09
feat: add env var to specify config path directly

TECHNOFAB11:
#### :zap: Summary

Adds an env var "LEFTHOOK_CONFIG" to specify the config file, can be relative or absolute.

- [x] Check locally
- [x] Add tests
- [ ] Add documentation (where to put that? the entry in --help might be enough?)

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -57,7 +57,28 @@ func (err ConfigNotFoundError) Error() string {
 	return err.message
 }
 
+func loadConfig(k *koanf.Koanf, filesystem afero.Fs, config string) error {
+	extension := filepath.Ext(config)
+	log.Debug("loading config: ", config)
+	if err := k.Load(kfs.Provider(newIOFS(filesystem), config), parsers[extension], mergeJobsOption); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func loadOne(k *koanf.Koanf, filesystem afero.Fs, root string, names []string) error {
+	configPathOverride := os.Getenv("LEFTHOOK_CONFIG")
+	if configPathOverride != "" {
+		if !filepath.IsAbs(configPathOverride) {
+			configPathOverride = filepath.Join(root, configPathOverride)
+		}
+		if ok, _ := afero.Exists(filesystem, configPathOverride); !ok {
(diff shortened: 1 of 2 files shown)


mrexox:
Could you share the problem you’re trying to solve, please?

TECHNOFAB11:
Sure, integration with Nix. Generating the config file from Nix, then wrapping lefthook to use that (eg `lefthook --config /nix/store/....yml $@`). Otherwise I'd have to generate the config file from Nix and always have to keep it in sync and in the git repo, which isn't great

mrexox:
@TECHNOFAB11 I see, so maybe a ENV variable would be a better choice? Lefthook can be run from within a git hook and there you can’t pass a `--config` parameter. ENV variable would be easier to tweak. WDYT?

TECHNOFAB11:
@mrexox I generate the git hook files myself so both a cli arg or env var would be fine. 
Would you rather just have the env var or have both, by defaulting the cli arg to the env var?

mrexox:
@TECHNOFAB11 I’d better use just the ENV var because it’s more flexible and requires less code to implement. Also lefthook is kind of ”hidden” tool which runs implicitly by Git hooks, so I assume that ENV var would have more users.

If there’s LEFTHOOK_CONFIG var set, let’s think that it overwrites any other configs (even lefthook-local.yml). 

By the way, do you need help implementing this?

TECHNOFAB11:
@mrexox sounds good, will modify the PR later today to use an env var instead 👍

cmd/root.go
```diff
@@ -11,7 +13,8 @@ import (
 
 func newRootCmd() *cobra.Command {
 	options := lefthook.Options{
-		Fs: afero.NewOsFs(),
+		Fs:         afero.NewOsFs(),
+		ConfigPath: os.Getenv("LEFTHOOK_CONFIG"),
```
mrexox, line 17: Let’s move `os.Getenv` to the `load.go`? This way we won’d have to change function signature. For testing you can use `t.Setenv`

mrexox (approved):

TECHNOFAB11:
Conflicts resolved :+1:

TECHNOFAB11:
Seems like a line got lost when fixing the conflicts, whoops, fixed :D

PR #1086, opened 2025-08-04
Improve wording of `run`, `files`, and `files-global` config descriptions, document that the `sh` shell is used

ItsHarper:
#### :zap: Summary

I cleaned up the wording of the first line of the `run`, `files`, and `files-global` config entry descriptions a little bit, and added the information that the `sh` shell is used.

I removed the word "git" since my understanding is that there is no requirement that the provided command be a git command.

- [ ] Check locally
- [ ] Add tests
- [x] Add documentation


mrexox:
Thank you! This way it makes more sense!

PR #1089, opened 2025-08-07
docs: update pnpm installation note

skoch13:
There is a way not to turn off `side-effects-cache` completely, so I provided updated guidelines


mrexox:
Oh, thank you for contribution!

PR #1090, opened 2025-08-08
fix: respect LEFTHOOK_CONFIG in lefthook install

TECHNOFAB11:
#### :zap: Summary

Just recently found out that `lefthook install` has it's own logic for checking if the config exists which I didn't notice when implementing support for `LEFTHOOK_CONFIG`. This adds the same logic for handling the config override to the install command, this way it doesn't create `lefthook.yml` when `LEFTHOOK_CONFIG` is set.

It actually currently still loads the correct config from the env var, just the install command didn't correctly find it and always created a sample config which was ignored. 

- [x] Check locally
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/internal/command/install.go
+++ b/internal/command/install.go
@@ -89,6 +89,17 @@ func (l *Lefthook) configExists(path string) bool {
 }
 
 func (l *Lefthook) findMainConfig(path string) (string, error) {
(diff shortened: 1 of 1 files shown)


mrexox:
Thank you for noticing and fixing this bug!

mrexox:
I will release this change tomorrow :+1:

Issue #1087, opened 2025-08-04
RFE: improved file type selection


scop:
### :zap: Summary

It is currently quite a chore to limit lefthook jobs to certain types of files when they cannot be adequately identified by globs. For example, projects may have files without extensions that are  written in shell, python, whatnot.

file_types exists, but is currently too simplistic for many use cases. There could be a native way to match types of files more granularly.

For example, I can  imagine lefthook making use of https://github.com/gabriel-vasile/mimetype, and file_types extended so that one could write mime types in it, in addition to what's currently possible. Lookup using mimetype could for perf reasons be made so that it is triggered only if file_types contains entries that look like mime types (maintype/subtype i.e. contain a slash).

mimetype supported files ref: https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md

Other libs with much larger coverage of various files besides mimetype exist, but they typically require libmagic which is a bit of a build/deployment issue.

### Value

Not having to maintain a big list of glob entries with individual filenames in it for affected projects.

### Behavior and configuration changes

file_types to support mime type matching, for example

```yaml
file_types:
  - text
  - executable
  - text/x-python
```

```yaml
file_types:
  - text/x-perl
```

mrexox:
Hey! Thank you for suggesting this feature. It totally makes sense, but I'm curious what kind of files are really wanted? I'd like to limit them because I don't think someone would use lefthook to work with video/audio files, it's just an extra functionality that needs to be supported if added. I guess only `text/*` checks are important, right? WDYT if I add support only for `text/*` types?

scop:
I think text types are most important, if it helps the implementation some way. But without knowing the full context, I suppose adding such a limitation could mean _additional_ code and complexity instead of just adding support for everything let's say the mentioned `mimetype` package supports. Also not sure what kind of extra functionality you mention would be needed, if the extent is just to be able to limit files to process.

But as such, I don't think limiting to `text/*` would be necessarily great in principle. For example JSON's official media type is `application/json` and JSON is definitely something one may want to trigger Lefthook jobs on.

As examples of hypothetical jobs processing non-text files, one might want to add a Lefthook job that checks that PNG files are optimized or JPEG files don't contain unwanted metadata, etc.

[...]

PR #1092, opened 2025-08-11
feat: add more filetype filters

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/1087

**:zap: Summary**

Add support for a few mime types in `file_types` filter

### DIFF
--- a/internal/run/filters/filters.go
+++ b/internal/run/filters/filters.go
@@ -250,7 +250,7 @@ func parseFileTypeFilter(types []string) fileTypeFilter {
 			filter.simpleTypes |= typeBinary
 		case t == "text":
 			filter.simpleTypes |= typeText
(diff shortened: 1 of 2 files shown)


internal/run/filters/filters.go
```diff
@@ -208,29 +233,48 @@ func byType(fs afero.Fs, vs []string, types []string) []string {
...
+			"text/x-python",
+			"text/rtf",
+			"text/x-tcl",
+			"text/csv",
+			"text/tab-separated-values",
+			"text/vcard",
+			"text/calendar",
+			"text/vtt",
+			"text/x-shellscript",
+			"text/x-sh",
```
scop, line 269: If the intent is to support text/* types, maybe this could be refactored to do `strings.HasPrefix("text/")` so that when new such types are added to mimetype, they would not need to be separately added to lefthook?

PR #1104, opened 2025-09-07
fix: Set `--force` for `git add` command

michael-pplx:
#### :zap: Summary

By default, `git` will fail if you try to stage a file that matches a `.gitignore` pattern. For _reasons_, a repo might have an ignored file already checked-in to the repo. These might be files that are checked by Lefthook, which means that if a job has `stage_fixed: true`, we don't want Lefthook to fail when it tries to stage them.

The relevant change is in `internal/git/repository.go`.

- [✅] Check locally
- [❌] Add tests
   Not sure how best to do this.
- [❌] Add documentation
   While this is technically a behavior change, I consider it to be a fix and doesn't need to be documented anywhere but the changelog.


mrexox:
Hey! Thank you for contribution! Could you please revert the changes of a version? I'll do it myself before preparing the release.

mrexox:
I'll merge the PR but will release it a bit later because I want a few other PRs merged before cutting a new version.

michael-pplx:
Sounds good, thanks!

PR #1112, opened 2025-09-09
feat: Add --no-stash-unstaged flag

michael-pplx:
Sorry for not opening an issue to discuss first. The PR seemed to trivial so it just seemed easier to open the conversation with that.

#### :zap: Summary

We have a script that wraps Lefthook for folks who want to run these jobs manually. However, we'd like to have a way to run pre-commit hooks over whatever local file modifications one happens to have, even if they aren't staged yet or are partially staged. This will also help us avoid a particular scenario where:
- We have an incorrectly formatted file staged.
- We make an additional unstaged change to the file.
- We run `lefthook run pre-commit` with a job that has `stage_fixed: true`.
  - Lefthook stashes the unstaged change.
  - A pre-commit job formats the file.
  - Lefthook attempts to unstash the unstaged change, but encounters a conflict.

By having a `--no-stash-unstaged` flag, our wrapper script can just run jobs over files in their current state and "just work".

Please let me know if this use-case is compatible with Lefthook's intended use and whether this is the right solution.

- [⏳] Check locally - In progress
- [❌] Add tests - I checked `internal/run/run_test.go` but it seems like it doesn't have existing logic for working with unstaged files, otherwise it would be good if there was a way to test whether `git stash` is called.
- [❓] Add documentation - I added a `--help` string, but I'm not sure where else to update. I just grepped for `no-auto-install` to find potential places to update.


internal/git/repository.go
```diff
@@ -245,6 +246,22 @@ func (r *Repository) PartiallyStagedFiles() ([]string, error) {
...
 
+func (r *Repository) ExcludePartiallyStagedFiles(files []string) ([]string, error) {
+	partiallyStagedFiles, err := r.PartiallyStagedFiles()
+	if err != nil {
+		return nil, err
+	}
+
+	onlyStagedFiles := make([]string, 0)
+	for _, file := range files {
+		if !slices.Contains(partiallyStagedFiles, file) {
```
michael-pplx, line 257: This could be inefficient if this list is long, leading to `O(n^2)` performance. We might be able to perform this in `O(n)` if we can assume that both `files` and `partiallyStagedFiles` have the same sort order. If not, then this could be speed up by using a `map[string]bool` like a Set.

mrexox:
Hey! Thank you for opening the PR, I will consider your use case, and will come with the review and ideas tomorrow.

So, basically, you'd like an CLI option to disable `stage_fixed` setting (and all related logic), right?

michael-pplx:
> Hey! Thank you for opening the PR, I will consider your use case, and will come with the review and ideas tomorrow.
> 
> So, basically, you'd like an CLI option to disable `stage_fixed` setting (and all related logic), right?

Yep, basically a way to achieve this behavior: https://github.com/evilmartians/lefthook/compare/master...michael-pplx:lefthook:no_stage_fixed

Issue #1084, opened 2025-07-28
lefthook throws `git add` warning/errors when stage_fixed is enabled


iven:
### :wrench: Summary

When I make a commit, lefthook sometimes throws following warning, even if no file is changed/formatted by lefthook:

```
> git add web/.cursorrules web/.github/instructions/basic.instructions.md ... <a long long list of file paths>
The following paths are ignored by one of your .gitignore files:
web/.vscode
hint: Use -f if you really want to add them.
hint: Disable this message with "git config set advice.addIgnoredFile false"
```

Seems lefthook is trying to add all files matching the glob, including ignored ones, regardless of whether the files have been modified.

Sometimes a different error occurs:

```
fatal: Unable to create 'project_path/.git/index.lock': File exists.
```

I guess it's because the `parallel` config is set to `true`, and multiple `git add` commands run at the same time.

### Lefthook version

1.12.2

### Steps to reproduce

1. Add `.vscode/` to `.gitignore`, but force add `.vscode/.gitkeep` file to the tree, and commit.
2. Enabled `parallel` and `stage_fixed` options.
3. Add several hooks.
4. Add a glob `*` to one of these hooks(so it matches `.vscode/`, and change another file that matching the glob.
5. `git commit` again, lefthook will report the ignored file error, and maybe the index.lock error.

### Expected results

No warning/error reported.

### Actual results

See logs above.

### Possible Solution

Only `git add` the files modified by lefthook. Only add once when `parallel: true` is set.

olivier-lacroix:
I have encountered similar contention.

Where jobs are run in parallel, I don't think lefthook can reliably determine which job modified which file. I am not sure having `stage_fixed` as a job parameter is desirable. Maybe it should be a hook setting instead?

mrexox:
It makes sense. I think about adding a mutex to prevent concurrent `git add`. 

`stage_fixed` respects the filters of the every job, so if you have 2 jobs configured – one for fixing `.md` and the other for `.js` files – each one will only stage the files it modifies. That's why this option was added to the jobs level.

But it may make sense adding a hook-level wrapping command, something like this:

```yml
after:
  - git add {staged_files}
```

But I'm not sure this is what people want 🤔

iven:
I'd prefer the way of the python pre-commit project: fail when any file changes, and let user check the diff and commit again.

olivier-lacroix:
[...]

PR #1115, opened 2025-09-11
fix: add mutex to prevent concurrent git adds

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/1084

### Context

For lots of jobs with `stage_fixed: true` lefthook may run `git add` concurrently and fail because Git creates a lockfile for such commands.

### Changes

Add a mutex to make `git add` calls sequential.

### DIFF
--- a/internal/git/repository.go
+++ b/internal/git/repository.go
@@ -53,6 +53,9 @@ var (
 
 // Repository represents a git repository.
 type Repository struct {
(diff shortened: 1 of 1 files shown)


ndbroadbent:
Amazing, I just came here to open an issue about this. Thanks so much for fixing it!

mrexox:
I have to say that it doesn't prevent locking issue if you do `git add ...` manually in some of the jobs which are being run in parallel ☝️

Issue #1125, opened 2025-09-15
fail_on_changes not working if the changed file is already committed


iven:
### Description

If a file is already committed(not any changes in git status), and lefthook job modifies this file, the `fail_on_changes` option will not be working.

In CI environment, all user modified files are already committed, so `fail_on_changes` will not work on CI.

### `lefthook.yml`

```
pre-commit:
  fail_on_changes: ci
  jobs:
    - name: test-job
      run: echo 123 >> README.md
```

### Commands to reproduce

```bash
$ git status -sb  # <- README.md not changed by user
## dev
 M lefthook.yml

$ CI=true lefthook run pre-commit --all-files --jobs test-job  # <- not failing
╭───────────────────────────────────────╮
│ 🥊 lefthook v1.13.0  hook: pre-commit │
╰───────────────────────────────────────╯
┃  test-job ❯

  ────────────────────────────────────
summary: (done in 0.01 seconds)
✔️ test-job (0.00 seconds)

$ git status -sb  # <- README.md changed by lefthook
## dev
 M README.md
 M lefthook.yml
```

### Lefthook version

1.13.0

olivier-lacroix:
@iven thanks for the report. [#1128](https://github.com/evilmartians/lefthook/pull/1128) should fix it

PR #1128, opened 2025-09-16
Fix: Clear cache before second call to ChangeSet

olivier-lacroix:
Closes https://github.com/evilmartians/lefthook/issues/1125

The `ChangeSet` method was relying on a cached git status 🙄

### DIFF
--- a/internal/run/controller.go
+++ b/internal/run/controller.go
@@ -237,6 +237,7 @@ func (c *Controller) preHook() {
 
 func (c *Controller) postHook() error {
 	if c.FailOnChanges {
+		c.Repo.Setup() // Reset the cache to get an up-to-date git status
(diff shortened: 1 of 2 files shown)


mrexox:
@olivier-lacroix thank you for noticing and preparing the fix. But I think calling `Setup()` is too much here. I've prepared a similar fix but with less git commands made: https://github.com/evilmartians/lefthook/pull/1129

olivier-lacroix:
okay, fair enough @mrexox . I liked the cleanliness of 'just' resetting the cache. Could we reset only `statusShortOnce` maybe?

mrexox:
I've used `statusShort` (without Once) for Changeset, so that it's always being called, avoiding the cache.

PR #1155, opened 2025-09-29
refactor!: use another cli framework

mrexox:
### Context

Maintaining cobra becomes hard, so I decided to switch to another, more declarative CLI framework and drop a few deprecated options. This will be released with the major version upgrade. Basic options won't change.

### Changes

- [x] Rewrite all commands with urfave/cli
- [x] Update the usage/docs
- [x] Add auto-completions

### ❗ Critical changes

- `lefthook run --jobs` replaced with `--job` 
- `lefthook run --commands` replaced with `--command` 
- `lefthook run --tags` replaced with `--tag` 
- `lefthook run --files` was dropped, use `--file a --file b`
- `lefthook dump --toml` was dropped, use `lefthook dump --format toml`
- `lefthook dump --json` was dropped, use `lefthook dump --format json`

### DIFF
--- a/docs/mdbook/usage/commands/run.md
+++ b/docs/mdbook/usage/commands/run.md
@@ -32,10 +32,10 @@ $ lefthook run pre-commit # will run pre-commit hook (`yarn lint --fix`)
 
 ### Run specific jobs
 
-You can specify which jobs to run (also `--commands`, `--scripts`, and `--tags` supported).
+You can specify which jobs to run (also `--tag` supported).
 
 ```bash
-$ lefthook run pre-commit --jobs lints --tags checks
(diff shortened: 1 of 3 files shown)


scop:
Looks like as part of this, shell completions for hook, command, and job names are no more. That would be a big loss, any plans to add them back?

mrexox:
Agree, I found that completions for flags is still under development: https://github.com/urfave/cli/issues/1905 (doesn't seem to be actively developed).

Currently only completions for hook are working `lefthook run <tab>`. I will try poking the issue of `urfave/cli` and maybe will find a workaround to make completions for flags work.

Issue #1173, opened 2025-10-24
Git lock contention


siler:
### Description

I see this error sometimes when running parallel jobs that interact with git:

```
  > git add --force go/services/molasses-be/internal/integration_test/integration_test.go
│    fatal: Unable to create './src/code/.git/index.lock': File exists.   
│    Another git process seems to be running in this repository, e.g.                     
│    an editor opened by 'git commit'. Please make sure all processes                     
│    are terminated then try again. If it still fails, a git process                      
│    may have crashed in this repository earlier:                                         
│    remove the file manually to continue.                                                
│                                                                                         
│  Couldn't stage fixed files:error in batch 0: exit status 128
```

When this happens everything actually succeeds unless something else triggers a failure, which may mean lefthook is not handling git failures well either.

### `lefthook.yml`

```
pre-commit:
  parallel: true
  jobs:
    - name: 'go'
      glob: '*.go'
      env:
        PATH: '$CODE_ROOT/scripts/hooks:$PATH'
        GOROOT: '$CODE_ROOT/bazel-code/external/go_sdk'
      group:
        piped: true
        jobs:
          - name: 'format'
            run: 'golangci_fmt.sh {staged_files}'
            stage_fixed: true
          - name: 'lint'
            run: 'golangci_lint.sh {staged_files}'
            stage_fixed: true
    - name: 'gazelle'
      glob:
        - '*.{go,proto}'
        - '*BUILD.bazel'
      env:
        PATH: '$CODE_ROOT/scripts/hooks:$PATH'
      run: 'gazelle.sh {staged_files}'
      stage_fixed: true
```

### Commands to reproduce

It is fairly random, but happens when files are changed by the hook. Just a pre-commit hook.

### Lefthook version

1.13.4 04831c8bff0f7b0a6ade1b06afae4f1b2228e82e

### Possible solution

Perhaps we should check for a lock and if there is one retry every 100-200ms?

mrexox:
Hey @siler Thank you for filing a bug. Do you use scripts that call `git` commands or are they only changing the files?

siler:
They are only changing files, none of the scripts call git themselves as far as I know.

siler:
Now that I'm thinking about it I sometimes see lock contention when just running normal git commands. It's possible I have automated systems using git on my computer (like vscode extensions for example)

PR #1178, opened 2025-10-27
fix: lock before all git commands

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/1173

### Context

Adding a mutex on just `stage_fixed` didn't help. That's probably because Git uses locks not only for `git add`, but also for `git stash`, `git diff`, and many other subcommands. Although I couldn't discover what exactly is causing the problem, I think it can be caused by the concurrent calls to `git`. 

If it doesn't help the problem may not be just in lefthook, maybe other programs are using git while lefthook is running hooks. In this case we need an external synchronization mechanism or a retry logic.

### Changes

When it's `git` command, use the mutex.

### DIFF
--- a/internal/command/run_test.go
+++ b/internal/command/run_test.go
@@ -33,13 +33,11 @@ func TestRun(t *testing.T) {
 	configPath := filepath.Join(root, "lefthook.yml")
 
 	for i, tt := range [...]struct {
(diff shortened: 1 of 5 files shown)


siler:
Thanks! I'll give this a try.

PR #1180, opened 2025-10-31
feat: Add stream option for real-time output without requiring TTY

### LINKED ISSUE #1179: feat: Add stream option for real-time output without requiring TTY
### Description

Add a new `stream` configuration option that enables real-time log streaming for commands and scripts without requiring a TTY. This would provide an alternative to `interactive: true` that works in non-TTY environments.

### What problem it is solving?

Currently, using `interactive: true` in non-TTY environments (like VS Code's commit interface, GitHub Desktop, or CI/CD pipelines) fails with the error:

```
Couldn't enable TTY input: open /dev/tty: device not configured
```

This happens because `interactive: true` attempts to open `/dev/tty` for user input, which doesn't exist in these environments.

Users who want real-time streaming output (like test results or linter output) but don't need user input are forced to either:
1. Use `interactive: true` and deal with TTY errors in non-TTY environments
2. Accept buffered output (default behavior using pseudo-TTY)

### Proposed solution

Add a `stream: true` option that:
- Streams output in real-time without requiring TTY access
- Does not attempt to open `/dev/tty` for input
- Works in all environments (TTY and non-TTY)
- Provides the same real-time output visibility as `interactive` but without the input requirement

### Use case example

```yml
pre-commit:
  commands:
    tests:
      run: npm test
      stream: true
    lint:
      run: npm run lint
      stream: true
```

This would allow developers to see test and lint output in real-time when committing from VS Code, without encountering TTY errors.

jasonwbarnett:
Closes #1179

### Context

Currently, using `interactive: true` in non-TTY environments (like VS Code's commit interface, GitHub Desktop, or CI/CD pipelines) fails with the error:

```
Couldn't enable TTY input: open /dev/tty: device not configured
```

This happens because `interactive: true` attempts to open `/dev/tty` for user input, which doesn't exist in these environments. Users who want real-time streaming output but don't need user input have no good alternative.

### Changes

This PR adds a new `stream` configuration option that enables real-time log streaming without requiring a TTY:

- **Added `stream` field** to Command, Script, and Job config structs
- **Updated executor logic** in `exec_unix.go` and `exec_windows.go` to handle stream mode
  - Stream mode pipes stdout/stderr directly without attempting to open `/dev/tty`
  - Only `interactive: true` attempts TTY access; `stream: true` works in all environments
- **Updated JSON schemas** with the new `stream` property
- **Added comprehensive documentation** at `docs/mdbook/configuration/stream.md`

**Key differences:**
- `stream: true` - Real-time output, no TTY required, no user input
- `interactive: true` - Requires TTY for user input
- `use_stdin: true` - Passes stdin but doesn't control output streaming  
- Default (all false) - Uses pseudo-TTY for output buffering

**Example usage:**
```yml
pre-commit:
  commands:
    tests:
      run: npm test
      stream: true
    lint:
      run: npm run lint
      stream: true
```

This allows developers to see test and lint output in real-time when committing from VS Code or other non-TTY environments, without encountering TTY errors.


mrexox:
Hey! Thank you for preparing the PR. Have you seen the [`follow`](https://lefthook.dev/configuration/follow.html) option? It must connect directly to the STDOUT and print output as it occurs. It does pretty the same as `stream`

jasonwbarnett:
No, I had not seen follow 🫠 

I'll try it out today and let you know if I have issues. Thanks!

jasonwbarnett:
It would be nice to control this on a per job basis though instead of the entire hook.

mrexox:
Agree, but maybe it will require to change the order of jobs and delay printing the output of those without `follow: true`. But that's a valid feature request. Do you want to reopen the issue?

jasonwbarnett:
Re-opened! 😄 

Let me know if you think we need to make any changes to the source or docs, etc. Happy to work with you to get this in. Also, let me know if you think.

PR #1186, opened 2025-11-07
feat: fail_on_changes non-ci option

scop:
### Context

In some setups, not failing on changes in CI is desirable, while it is desirable to fail in non-CI. So this is essentially the opposite of `ci`. See docs in the PR for more info.

Caveat: this is **untested**. I would expect to be able to remove this block of config altogether if this option existed:
https://github.com/scop/pytekukko/blob/07074196c9fc73e7ca4d06ba4372fe727e0900c1/.config/lefthook.yaml#L22-L28

### Changes

Sprinkle `non-ci` around where `ci` was found.


mrexox:
Thank you for preparing the PR!

PR #1188, opened 2025-11-11
feat: add optional standard glob matcher (doublestar)

### LINKED ISSUE #1187: Glob pattern behavior differs from standard implementations, causing user confusion
## Summary

The current glob pattern behavior in lefthook, particularly with `**`, differs from standard glob implementations and causes confusion for users. While the documentation does mention this, the non-standard behavior creates friction for users coming from other tools and makes patterns harder to reason about.

## Current Behavior

In lefthook using gobwas/glob:
- `*.js` matches files at **all levels** (both root and nested): `file.js`, `folder/file.js`, `a/b/file.js`
- `**/*.js` matches files **only in nested directories** (1+ levels deep): `folder/file.js`, `a/b/file.js` 
- `**/*.js` does **NOT** match `file.js` at the root level

This means:
```yaml
glob: "*.js"  # matches both root and nested - unusual behavior!
```

## Expected/Standard Behavior

In most glob implementations (bash, zsh, .gitignore, pre-commit, lint-staged, etc.):
- `*.js` matches files **only at the root level**: `file.js`
- `**/*.js` matches files **at all levels** (0+ directories): `file.js`, `folder/file.js`, `a/b/c/file.js`

## Why This Matters

1. **Unexpected `*` behavior**: The `*` wildcard recursively matching subdirectories is very unusual and counterintuitive
2. **Inverted semantics**: `**` requiring at least one directory is the opposite of standard glob behavior
3. **User Expectations**: Users coming from other tools expect standard glob behavior and are surprised when patterns don't work as expected
4. **Learning Curve**: This adds an unnecessary learning curve and requires users to unlearn their existing glob knowledge
5. **Documentation Burden**: Even with documentation, users (and AI agents assisting users) frequently get this wrong
6. **Portability**: Patterns can't be easily copied from other tools like .gitignore, pre-commit configs, etc.

## Real-World Impact

I've personally experienced this confusion, and when working with AI coding assistants (Claude, GitHub Copilot, etc.), they consistently generate incorrect patterns because they're trained on standard glob behavior. This creates a feedback loop where:
1. AI suggests a pattern based on standard glob behavior
2. Pattern doesn't work as expected in lefthook
3. User has to debug and learn the non-standard behavior
4. This repeats for each new pattern

## Proposed Solution

I'd like to suggest one of the following approaches:

### Option 1: Switch to Standard Glob Behavior (Breaking Change)
Change glob behavior to match standards in the next major version. This would be a breaking change but would align with user expectations.

### Option 2: Add an Opt-in Configuration (My Implementation)
Add a `glob_matcher` configuration option to allow users to opt-in to standard behavior:
```yaml
glob_matcher: doublestar  # or "standard"

pre-commit:
  commands:
    lint:
      glob: "**/*.js"  # now matches files at any level including root
      run: eslint {staged_files}
```

I've actually implemented this approach and would be happy to open a PR. The implementation:
- Uses `github.com/bmatcuk/doublestar/v4` as an alternative matcher
- Maintains backward compatibility (default behavior unchanged)
- Allows opt-in via configuration
- Is fully tested with comparison tests showing both behaviors

### Option 3: Better Documentation + Migration Guide
If changing the behavior isn't feasible, perhaps we could:
- Make the non-standard behavior more prominent in docs
- Add migration examples for common use cases
- Include a "glob pattern translator" in the docs

## Community Feedback

I'd love to hear from other users if they've experienced similar confusion. I think this is a common pain point but may not be reported often because users assume they're doing something wrong rather than the tool having non-standard behavior.

## Context

I'm opening this issue with empathy and appreciation for lefthook - it's a fantastic tool and I use it daily. This isn't a criticism of the design decision, just feedback on a friction point I've encountered repeatedly. I'm happy to contribute a PR for any of the proposed solutions if the maintainers are open to addressing this.

jasonwbarnett:
Closes #1187

### Context

The current glob pattern behavior in lefthook differs from standard implementations in two ways:
1. `*` wildcard recursively matches subdirectories (very unusual)
2. `**` requires at least 1 directory level instead of the standard 0+

Current behavior (gobwas):
- `*.js` matches at **all levels**: `file.js`, `folder/file.js`, `a/b/file.js`
- `**/*.js` matches **only nested** (1+ levels): `folder/file.js`, `a/b/file.js`

Standard behavior (bash, zsh, .gitignore, pre-commit, lint-staged):
- `*.js` matches **only root level**: `file.js`
- `**/*.js` matches **all levels** (0+ levels): `file.js`, `folder/file.js`, `a/b/file.js`

As discussed in #1187, this non-standard behavior creates friction:
- Users must learn lefthook-specific glob patterns
- Patterns can't be easily copied from other tools
- AI coding assistants consistently generate incorrect patterns
- Documentation burden to explain the difference

This PR implements an opt-in solution that maintains full backward compatibility while allowing users to choose standard glob behavior.

### Changes

**Core Implementation:**
- Add `glob_matcher` config option with values: `gobwas` (default) or `doublestar`
- Implement dual matcher support in filters package using `github.com/bmatcuk/doublestar/v4`
- Refactor glob/exclude matching logic to reduce complexity (fixes nestif linter issues)
- Thread `GlobMatcher` through the entire execution pipeline

**Testing:**
- Add `TestByGlobDoublestar` with 5 test cases comparing both matchers
- Add `TestByExcludeDoublestar` with 4 test cases for exclusion patterns
- Update existing tests to include `globMatcher` parameter
- All tests pass including race condition testing

**Documentation:**
- New `docs/mdbook/configuration/glob_matcher.md` with complete reference
- Update `glob.md` to reference the new option
- Update `exclude.md` to note glob_matcher affects exclude patterns  
- Add entry to `SUMMARY.md` table of contents
- Create implementation guide for developers

**Behavior Comparison:**

Default (gobwas):
```yaml
glob: "*.js"
# Matches: file.js, folder/file.js, a/b/file.js (all levels - unusual!)

glob: "**/*.js"  
# Matches: folder/file.js, a/b/file.js (nested only)
# Does NOT match: file.js (root level)
```

With `glob_matcher: doublestar`:
```yaml
glob_matcher: doublestar

glob: "*.js"
# Matches: file.js (root level only - standard behavior)

glob: "**/*.js"
# Matches: file.js, folder/file.js, a/b/file.js (all levels - standard!)
```

**Files Changed:** 16 files
- 8 implementation files
- 1 test file  
- 4 documentation files
- 3 generated/schema files

**Backward Compatibility:**
- ✅ Default behavior unchanged
- ✅ Opt-in only via configuration
- ✅ No breaking changes
- ✅ All existing configs work as-is

This implementation provides a path forward for users who want standard glob behavior while respecting the existing behavior for current users.

### DIFF
--- a/docs/mdbook/SUMMARY.md
+++ b/docs/mdbook/SUMMARY.md
@@ -54,6 +54,7 @@
   - [`source_dir`](./configuration/source_dir.md)
   - [`source_dir_local`](./configuration/source_dir_local.md)
   - [`skip_lfs`](./configuration/skip_lfs.md)
+  - [`glob_matcher`](./configuration/glob_matcher.md)
   - [`templates`](./configuration/templates.md)
   - [{Git hook name}](./configuration/Hook.md)
     - [`files`](./configuration/files-global.md)
--- a/docs/mdbook/configuration/exclude.md
+++ b/docs/mdbook/configuration/exclude.md
@@ -2,6 +2,8 @@
 
 This option allows to setup a list of globs for files to be excluded in files template.
 
+> **Note:** The glob patterns used in `exclude` are affected by the [`glob_matcher`](./glob_matcher.md) setting. See the glob_matcher documentation for details on how `**` patterns behave.
+
 **Example**
 
 Run Rubocop on staged files with `.rb` extension except for `application.rb`, `routes.rb`, `rails_helper.rb`, and all Ruby files in `config/initializers/`.
--- a/docs/mdbook/configuration/glob.md
+++ b/docs/mdbook/configuration/glob.md
@@ -50,6 +50,8 @@ You'll need:
 glob: "src/*.js"
 ```
 
+Alternatively, you can opt-in to standard glob behavior by setting [`glob_matcher: doublestar`](./glob_matcher.md) at the top level of your configuration. With this setting, `**` will match 0 or more directories, making it consistent with most other glob implementations.
+
 ***Using `glob` without a files template in`run`***
 
 If you've specified `glob` but don't have a files template in [`run`](./run.md) option, lefthook will check `{staged_files}` for `pre-commit` hook and `{push_files}` for `pre-push` hook and apply filtering. If no files left, the command will be skipped.
--- a/docs/mdbook/configuration/glob_matcher.md
+++ b/docs/mdbook/configuration/glob_matcher.md
@@ -1,0 +1,85 @@
+## `glob_matcher`
+
+You can configure which glob matching engine lefthook uses to filter files. By default, lefthook uses `gobwas/glob`, but you can opt-in to use `doublestar` for standard glob behavior.
+
+**Values:**
+- `gobwas` (default): The current glob implementation
+- `doublestar`: Standard glob behavior where `**` matches 0 or more directories
+
+**Example:**
+
+```yml
+# lefthook.yml
+
+glob_matcher: doublestar
+
+pre-commit:
+  jobs:
+    - name: lint
+      run: yarn eslint {staged_files}
+      glob: "**/*.{js,ts}"
+```
+
+### Key Differences
+
+The main difference between the two matchers is how they handle `**`:
+
+#### Default behavior (`gobwas`)
+
+The `**` pattern matches **1 or more** directories:
+- `**/*.js` matches `folder/file.js`, `a/b/c/file.js`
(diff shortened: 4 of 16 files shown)


mrexox:
Thank you for preparing a PR! And huge thanks for keeping backward compatibility. I will review the PR in a few days :+1:

mrexox (approved):
All good, thanks again for preparing this PR!

PR #1213, opened 2025-12-04
fix: do not escape custom templates in command replacement

joevin-slq-docto:
Closes https://github.com/evilmartians/lefthook/issues/1214

## Summary

This PR fixes a regression introduced in v2.0.5 (https://github.com/evilmartians/lefthook/pull/1206) where custom templates defined in the `templates` config section are incorrectly shell-escaped, causing command execution failures.

## Problem Description

Custom templates (user-defined templates in the config) are being wrapped in shell quotes when replaced in commands, breaking shell command execution. This happens because the refactoring in https://github.com/evilmartians/lefthook/pull/1206 unified the handling of all templates but inadvertently applied shell escaping to custom templates which should only be applied to file path templates.

**`lefthook.yml` config file:**
```yaml
templates:
  use-mise: eval "$(mise env)"

pre-commit:
  jobs:
    - name: prettier
      run: |
        {use-mise}
        prettier --check {staged_files}
```

**Expected command (v2.0.4):**
```bash
eval "$(mise -C .git/info/lefthook-remotes/hooks-main env)" # template should be replaced with its value without quotes
prettier --check file1.js
```

**Actual command (v2.0.5):**
```bash
'eval "$(mise -C .git/info/lefthook-remotes/hooks-main env)"' # template is replaced but incorrectly wrapped in quotes
prettier --check file1.js
```

This results in:
```
sh: 1: eval "$(mise -C .git/info/lefthook-remotes/hooks-main env)": not found
```

### Verbose Output

With `LEFTHOOK_VERBOSE=1`, you can see:
```
[lefthook] files after escaping: 'eval "$(mise -C .git/info/lefthook-remotes/hooks-main env)"'
[lefthook] job: 'eval "$(mise -C .git/info/lefthook-remotes/hooks-main env)"'
echo "Running with mise"
```

## Root Cause

In commit 57b1094, the template replacement logic was refactored from `build_command.go` into a new `replacer/replacer.go` package.

**Before (v2.0.4)** - in `build_command.go`:
```go
// Custom templates replaced first with simple string replacement
for keyword, replacement := range b.opts.Templates {
    runString = strings.ReplaceAll(runString, "{"+keyword+"}", replacement)
}

// Then file templates processed and escaped
commands, files := replaceInChunks(runString, filesTemplates, maxlen)
```

**After (v2.0.5)** - in `replacer/replacer.go:132-182`:
```go
func (r Replacer) ReplaceAndSplit(command string, maxlen int) ([]string, []string) {
    // ...
    for template, entry := range r.cache {
        // ...
        entry.items = escapeFiles(entry.items)  // ❌ Escapes EVERYTHING
    }
    // ...
}
```

The new unified approach applies `escapeFiles()` to **all** templates (line 151), including custom templates. However, custom templates often contain shell commands that must be executed directly, not escaped.

## Solution

This PR modifies `ReplaceAndSplit` to only escape file templates (which represent file paths) and not custom templates (which represent arbitrary user-defined replacements).

**Fix in `replacer.go:139-152`:**
```go
allFiles := make([]string, 0)
for template, entry := range r.cache {
    if entry.cnt == 0 {
        continue
    }

    cnt += entry.cnt
    maxlen += entry.cnt * len(template)
    if _, ok := r.files[template]; ok {
        allFiles = append(allFiles, entry.items...)
        // Only escape file templates, not custom templates
        entry.items = escapeFiles(entry.items)
    }
}
```

The distinction is made using `r.files` (file templates like `{staged_files}`) vs `r.templates` (custom user-defined templates).

## Tests

Added comprehensive tests in `replacer_test.go` to verify:
1. Custom templates are **not** escaped
2. File templates are **still** escaped (preserving existing behavior)

All existing tests continue to pass.

## Regression

- **Introduced in:** v2.0.5 (commit 57b1094)
- **Affected versions:** v2.0.5+
- Breaks custom templates containing shell commands

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change log
 
+## 2.0.7 (2025-12-04)
+
+- fix: prefer using lefthook from the $PATH ([#1211](https://github.com/evilmartians/lefthook/pull/1211)) by [@joevin-sql-docto]()
+
 ## 2.0.6 (2025-12-03)
 
 - feat: save original executable location in hooks ([#1208](https://github.com/evilmartians/lefthook/pull/1208)) by [@mrexox]()
@@ -1060,3 +1064,4 @@ gem 'lefthook'
 [@jasonwbarnett]: https://github.com/jasonwbarnett
 [@scop]: https://github.com/scop
 [@franzramadhan]: https://github.com/franzramadhan
+[@joevin-sql-docto]: https://github.com/joevin-slq-docto
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ A Git hooks manager for Node.js, Ruby, Python and many other types of projects.
 With **Go** (>= 1.25):
 
 ```bash
-go install github.com/evilmartians/lefthook/v2@v2.0.6
+go install github.com/evilmartians/lefthook/v2@v2.0.7
 ```
 
 * or as a go tool
--- a/docs/mdbook/installation/go.md
+++ b/docs/mdbook/installation/go.md
@@ -5,7 +5,7 @@ The minimum Go version required is 1.25 and you can install
 - as global package
 
 ```bash
-go install github.com/evilmartians/lefthook/v2@v2.0.6
+go install github.com/evilmartians/lefthook/v2@v2.0.7
 ```
 
 - or as a go tool in your project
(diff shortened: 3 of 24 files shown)


mrexox (approved):
Oh, thank you for preparing the fix!

PR #1217, opened 2025-12-05
fix: preserve job order when using extends in remote configs

joevin-slq-docto:
## Summary

Fixes job execution order when using `extends` in remote configurations. Jobs from the base configuration now execute before jobs from the extending configuration, preserving dependency chains.

Closes https://github.com/evilmartians/lefthook/issues/1218

## Problem

When a remote config uses `extends` to load another config, the job execution order was reversed:
- Jobs from the **extending** file executed first
- Jobs from the **base** file executed last

This broke dependency chains. For example, if a base config sets up dependencies (like `mise install`) and an extending config adds additional checks (like `prettier`), the checks would run before dependencies were installed.

## Root Cause

The `mergeJobsSlice` function in `internal/config/load.go` processed jobs in the wrong order:
1. It added `dest` (existing/extending config) jobs first
2. Then added `src` (new/base config) jobs

This order worked for local overrides but was incorrect for `extends`, where the base should come first.

## Solution

This PR restructures `mergeJobsSlice` to establish a clear semantic convention:
- **`src` = base configuration** (should execute first)
- **`dest` = override/extension** (should execute second)

### Changes made:

1. **Refactored `mergeJobsSlice`** (`internal/config/load.go:510-595`):
   - Now processes `src` jobs first, then `dest` jobs
   - Updated merge logic for jobs with matching names
   - Adjusted `{cmd}` template replacement to work with the new order

2. **Updated `addHook` call** (`internal/config/load.go:373`):
   - Swapped parameter order to maintain semantic consistency
   - Main config (base) is now passed as `src`, local override as `dest`

## Testing

- Verified the fix resolves the issue described in #1218:
  ```bash
  lefthook dump
  ```
  Now shows correct order:
  ```yaml
  pre-commit:
    jobs:
      - name: setup mise      # Base config (first)
      - name: hooks
        group:
          jobs:
            - name: ggshield   # Base config job
            - name: prettier   # Extension config job
  ```

### DIFF
--- a/tests/integration/job_merging.txt
+++ b/tests/integration/job_merging.txt
@@ -108,20 +108,20 @@ pre-commit:
       group:
         jobs:
           - name: child
-            run: child named with glob
+            run: child named
             stage_fixed: true
           - run: 0 no-name
-          - run: 1 no-name
-          - run: 2 no-name
           - run: 3 no-name
+          - run: 2 no-name
+          - run: 1 no-name
     - name: echo
-      run: echo 2
+      run: echo 1
       glob:
         - "3"
       tags:
         - backend
       skip: true
     - run: lefthook.yml
-    - run: e1
-    - run: e2
     - run: e3
+    - run: e2
+    - run: e1


mrexox:
We discussed the issue and decided to go with [`setup`](https://lefthook.dev/configuration/setup/) setting instead of changing the existing order of loading the extends.

PR #1223, opened 2025-12-05
chore: upgrade golangci-lint to 2.7.1, add godoclint

scop:
### Context

Better linting coverage.

### Changes

Upgrade golangci-lint 2.7.1, add the godoclint linter, address cropped up findings.


mrexox:
Thank you for taking care of linters!

Issue #1171, opened 2025-10-22
Ability to show diff on fail_on_changes


scop:
### Description

`fail_on_changes` is great, but it lacks the possibility to show diffs about what has actually changed. This would be especially important in CI because it's not easy to get hold of that environment and figure it out manually.

pre-commit has the `--show-diff-on-failure` option to cater to this.

My personal idea of the perfect behavior is, that if the tool detects changed files and fails due to them, it should show the diff of the tree iff in CI (i.e. _not_ show the diff if not in CI, because one can easily figure out locally if need be, and the diff output may wreak havoc in the tool doing the commit/showing its output).
My opinionated configs also set `fail_on_changes: always`, so the above behavior should not trigger only when `fail_on_changes: ci`. 

I appreciate the above might not work for everyone, so maybe the above described behavior could be the default, and another option to explicitly toggle showing of the diff (on or off) should be added (`--[no-]show-failure-diff`?).

Or if there are backwards compatibility considerations, maybe the diff would never be shown by default, but would require passing the `--show-failure-diff`.

### What problem it is solving?

Better understanding why fail_on_changes kicks in, especially/typically in CI.

iven:
+1 for this, no diff when fail_on_changes makes it not that useful.

scop:
Suggested implementation in #1227.

PR #1227, opened 2025-12-06
feat: ability to show diff when failing on changes

scop:
Closes https://github.com/evilmartians/lefthook/issues/1171

~Draft: untested.~ Tested now.

### Context

See above issue for discussion.

### Changes

Add a cmdline flag and config option for controlling the behavior, default to on in CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,14 +71,18 @@ jobs:
           name: dist
       - run: tar -xvf dist.tar
 
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Update npm
+        run: npm install -g npm@latest
+
       - name: Publish to NPM
         env:
           NPM_API_KEY: ${{ secrets.NPM_API_KEY }}
         run: |
-          cat << EOF > ~/.npmrc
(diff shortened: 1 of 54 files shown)


internal/run/controller/guard.go
```diff
@@ -119,3 +123,35 @@ func (g *guard) after() error {
 
 	return nil
 }
+
+func (g *guard) changesetDiff(changesetAfter map[string]string) {
```
scop: I don't particularly like the interface of the method here, I think I would have rather had it return the diff and an error, and checked `g.failOnChangesDiff` at the call site. However exposing more `if`s in `after()` runs afoul with the `nestif` linter. But nevermind, I can live with this :)

scop:
> What do you think about always showing diff on fail changes?

As noted in #1171, talking about the default: "i.e. _not_ show the diff if not in CI, because one can easily figure out locally if need be, and the diff output may wreak havoc in the tool doing the commit/showing its output"

For example, if an IDE shows the Lefthook output in a pop-up dialog, or something like https://git-cola.github.io shows it, having the diff in it may become quite a mess. (Even the usual "fancy" Lefthook output we have tends to do that.) And the diff may be huge, etc etc.

> I can work on diff styling later to display it not as the command output, but as lefthook output (with leading | ).

FWIW, I would personally prefer it to be shown as possibly colorized but otherwise unstyled plain diff.

PR #1230, opened 2025-12-09
feat: more rudimentary shell completions

scop:
Refs https://github.com/evilmartians/lefthook/pull/1155#issuecomment-3424871549

### Context

Since Lefthook 2.0.0, shell completions have regressed significantly due to change of the CLI library.

### Changes

Bring back more hook completions, as well as flag completions. Completions in general continue to be worse than they were in 1.x (lack of context), but I suppose this is borderline as far as they can be taken them without landing significant custom work (which would be better placed in the CLI lib).

### DIFF
--- a/cmd/add.go
+++ b/cmd/add.go
@@ -47,5 +47,9 @@ func add() *cli.Command {
 			args.Hook = cmd.Args().Get(0)
 			return l.Add(ctx, args)
 		},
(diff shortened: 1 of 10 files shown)


mrexox:
Thank you for adding this! Yes, after switching to a new CLI lib there's no completions for flag options, like `--job` and `--command`. But with flags it already feels much better!

Issue #1232, opened 2025-12-11
Command isn't skipped if files result is empty


whoisahmed:
### Description
[As per the docs](https://lefthook.dev/configuration/files.html#files), I was expecting the command not run if the `files` value is empty, but it still runs.

### `lefthook.yml`
```
    codeowners:
      tags: backend frontend
      files: git diff --name-only --diff-filter=AR --staged
      run: .lefthook/pre_commit/codeowners.sh {files}
```

### Lefthook version

lefthook version 2.0.9 

### Possible solution

Shouldnt run

ValentinGurkov:
I think I am observing the same but with `glob`

whoisahmed:
I think the last functional version is `2.0.2`

mrexox:
Thank you for reporting this bug. Indeed the behavior has changed since 2.0.2. I'll release the fix ASAP.

sanmai-NL:
@mrexox The issue is still present with `{all_files}` in 2.0.13.

sanmai-NL:
@mrexox It seems not fixed in general:

```yaml
    rustfmt:
      run: cargo fmt --all -- -- "{all_files}"
      file_types:
        - text
      glob:
        - '*.rs'
```

```
│  [lefthook] job: cargo fmt --all -- --
│  [lefthook] run: cargo fmt --all -- --
┃  rustfmt ❯ 

`cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/Users/myuser/devel/bla/_template` or any parent directory

This utility formats all bin and lib files of the current crate using rustfmt.

Usage: cargo fmt [OPTIONS] [-- <rustfmt_options>...]

Arguments:
  [rustfmt_options]...  Options passed to rustfmt

Options:
  -q, --quiet
          No output printed to stdout
  -v, --verbose
          Use verbose output
      --version
          Print rustfmt version and exit
  -p, --package <package>...
          Specify package to format
      --manifest-path <manifest-path>
          Specify path to Cargo.toml
      --message-format <message-format>
          Specify message-format: short|json|human
      --all
          Format all packages, and also their local path-based dependencies
      --check
          Run rustfmt in check mode
  -h, --help
          Print help


exit status 1                                      
  ────────────────────────────────────
summary: (done in 0.15 seconds)       
✗ rustfmt (0.15 seconds)
```

sanmai-NL:
And I conform that v2.0.2 does work, in the same context.

[...]

PR #1233, opened 2025-12-12
fix: skip if empty files template

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/1232

### Context

When `{files}` template resolves empty list, the command execution must be skipped. Even if the `{files}` template is in the `run` string.

### Changes

Check not only if files were cached, but also if there were actually any file.

### DIFF
--- a/internal/run/controller/command/build_command.go
+++ b/internal/run/controller/command/build_command.go
@@ -55,7 +55,7 @@ func (b *Builder) buildCommand(params *JobParams) ([]string, []string, error) {
 		}
 
 		if len(files) == 0 {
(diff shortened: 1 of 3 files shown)


ValentinGurkov:
@mrexox Sorry to ping you like that, the NPM publishing failed

mrexox:
Ok, probably will have to figure this out before publishing to NPM: https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/
Will take a few minutes/hours

mrexox:
All good now

mrexox:
Thank you for pinging me!

PR #1236, opened 2025-12-14
fix(git): make short status parser more robust

scop:
Ref https://git-scm.com/docs/git-status#_short_format

### Context

The current git short status parser has a number of problems with filenames containing spaces and other cases for which the "plain" porcelain format outputs as quoted (the surrounding quotes are not removed, nor are the backslash escaped characters within quotes unescaled), as well as pathological cases like ones containing ` -> ` (which are mistreated as renames, losing the latter half of the filename), etc.

### Changes

Switch to the NUL separated -z porcelain format recommended for machine parsing to make it easier to avoid hitting the mentioned issues. 

A downside is that parsing this format is stateful and we need to "know" the cases for which two filenames are output, currently copy (C) and rename (R).

Caveat: I have not found a way to produce a "C" state file locally, so I'm just assuming that it will produce two file entries like "R" does, based on https://git-scm.com/docs/git-status#_output
> where ORIG_PATH is where the renamed/copied contents came from

### DIFF
--- a/internal/git/repository.go
+++ b/internal/git/repository.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


mrexox:
Thank you! This change totally makes sense and I'm thankful for new tests!

PR #1246, opened 2025-12-15
fix: don't install custom hooks to hooks dir

scop:
### Context

There is no benefit to my knowledge from installing custom hooks to the git hooks dir. If that's true, we could avoid doing that and keep it a bit cleaner.

### Changes

Install only "real" hooks, rename existing "non-real" ones we may have installed earlier to `.old` on sync.


mrexox:
Oh, thank you for making this change!

PR #1250, opened 2025-12-18
fix: allow custom hooks in JSON schema by updating generator

### LINKED ISSUE #1249: JSON Schema does not allow custom hooks invoked by the run command
### Description
The JSON schema for lefthook doesn't allow custom hooks meant to be invoked using the `lefthook run` command. this is because `additionalProperties` in the schema is set to false.
https://github.com/evilmartians/lefthook/blob/009da6884b03e65c7d920ff0fd8ea53a49de6929/schema.json#L624

### `lefthook.yml`

Config file from the documentation page https://lefthook.dev/usage/commands/run.html

### Commands to reproduce

- git init in an empty directory
- create `lefthook.yml` from https://lefthook.dev/usage/commands/run.html
- run `lefthook validate`

### Lefthook version

`2.0.12`
latest at the time of writing

### Possible solution

Update JSON Schema to allow additional custom hooks
```json
    "additionalProperties": {
        "$ref": "#/$defs/Hook"
    },
```

jeonghoon11:
Closes #1249 

### Context
The current JSON schema for Lefthook sets `additionalProperties: false` at the root level. This prevents users from defining custom hooks that are intended to be executed via the `lefthook run <custom-hook>` command. When a custom hook is added to `lefthook.yml`, the validation fails with a message stating that no values are allowed for that key.

This PR resolves the issue by allowing the root object to accept additional properties, provided they conform to the `Hook` definition.

### Changes
Schema Generator Update: Modified `gen/jsonschema.go` to programmatically assign the `Hook` definition to the root-level `AdditionalProperties`. This ensures that any custom key in the configuration is validated against the standard hook schema.

Schema Re-generation: Updated both the public-facing `schema.json` and the embedded `internal/config/jsonschema.json` by running the updated generator.

Manual Validation: Verified the fix by running `./lefthook validate` with a configuration file containing a custom hook, which now returns "All good".

### DIFF
--- a/gen/jsonschema.go
+++ b/gen/jsonschema.go
@@ -54,6 +54,9 @@ func main() {
 		return []reflect.StructField{}
 	}
 	schema := r.Reflect(&config.Config{})
+	if hookDef, ok := schema.Definitions["Hook"]; ok {
+		schema.AdditionalProperties = hookDef
+	}
 	schema.ID = "https://json.schemastore.org/lefthook.json"
 	schema.Comments = "Last updated on " + time.Now().Format("2006.01.02") + "."
 	dumped, err := json.MarshalIndent(schema, "", "  ")
--- a/internal/config/jsonschema.json
+++ b/internal/config/jsonschema.json
@@ -442,7 +442,7 @@
       "type": "object"
     }
   },
-  "$comment": "Last updated on 2025.12.10.",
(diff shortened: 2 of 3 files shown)


mrexox:
Thank you for finding the way to add custom hooks to jsonschema!

PR #1251, opened 2025-12-18
Fix issue with restoring stashed code with partially edited files

### LINKED ISSUE #1216: Unstaged changes not restored when fail_on_changes is on
### Description

When fail_on_changes is on, some unstaged changes were stashed away, and there were some changes made by tools, the unstaged changes are not brought back. The offending too early return is

https://github.com/evilmartians/lefthook/blob/22171e0519200d2ca825e4165bb3c197029bdcad/internal/run/controller/guard.go#L102

I feel this is a serious problem, because the changes that were stashed away could be result of substantial work, and the user is not notified about this at all. If they do not take manual action to resolve the situation, it can result in data loss.  

I think pre-commit does this the right way: it always attempts to restore stashed changes. If there are conflicts due to some changes having appeared, it rolls back those changes and retries restoring the stashed ones. This is fine, because the appeared changes are very likely made by tools and can be easily redone, which might not hold for the changes that were stashed.

https://github.com/pre-commit/pre-commit/blob/98ccafa3ce42b846b9a9be9ed73fbbec7415496d/pre_commit/staged_files_only.py#L84-L98

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true
```

### Lefthook version

2.0.7

### Possible solution

Always attempt to restore stashed changes, no matter if fail_on_changes is on or off.

If there are problems restoring, revert changes made to the affected files (presumably by tools), and try restoring stashed again.

Output a warning if the above retry triggers.

tuchfarber:
Closes #1216

### Context

https://github.com/evilmartians/lefthook/issues/1216 happens because lefthook compares the pre-stash changeset with a post-stash changeset, meaning if you have `fail_on_changes` enabled and a partially updated file, it will fail.

### Changes

This PR adds a copy of the `changesetBefore` code after the stashing code so if we make it to stashing (i.e. it doesn't exit earlier) we reset the `changesetBefore` to be accurate.

NOTE: I'm not 100% confident on the test addition. I verified it fails in the current code and passes in the new code, and I've built the binary and manually tested as well. I attempted to match existing code/style.

### DIFF
--- a/internal/run/controller/guard.go
+++ b/internal/run/controller/guard.go
@@ -106,11 +106,19 @@ func (g *guard) before() {
 }
 
 func (g *guard) after() error {
-	changesetAfter, err := g.git.Changeset()
-	if err != nil {
-		log.Warnf("Couldn't get changeset: %s\n", err)
+	// Only get changeset if we need it for failOnChanges check
+	var changesetAfter map[string]string
+	var isFailingOnChanges bool
+
+	if g.failOnChanges {
+		var err error
+		changesetAfter, err = g.git.Changeset()
+		if err != nil {
+			log.Warnf("Couldn't get changeset: %s\n", err)
+			changesetAfter = make(map[string]string)
+		}
+		isFailingOnChanges = !maps.Equal(g.changesetBefore, changesetAfter)
 	}
-	isFailingOnChanges := g.failOnChanges && !maps.Equal(g.changesetBefore, changesetAfter)
 
 	if !g.didStash {
 		if isFailingOnChanges {
@@ -124,6 +132,12 @@ func (g *guard) after() error {
 		log.Warnf("Couldn't restore unstaged files: %s\n", err)
 		// If we can't restore the unstaged files, first roll back the changes
 		// introduced by the hook before trying to restore unstaged files again
+		// Get changeset only when needed for error recovery
+		changesetAfter, err := g.git.Changeset()
+		if err != nil {
+			log.Warnf("Couldn't get changeset: %s\n", err)
+			changesetAfter = make(map[string]string)
+		}
 		changed := g.getChangedFiles(changesetAfter)
 
 		log.Warnf("Couldn't restore unstaged files after hook changes, rolling back: %s\n", changed)
--- a/internal/run/controller/guard_test.go
+++ b/internal/run/controller/guard_test.go
@@ -115,6 +115,148 @@ func Test_guard_wrap(t *testing.T) {
 				{Command: "git stash drop --quiet 1", Output: ""},
 			},
 		},
+		"stashUnstagedChanges=true failOnChanges=true with partially staged and hook changes": {
+			stashUnstagedChanges: true,
+			failOnChanges:        true,
+			commands: []cmdtest.Out{
+				{Command: "git status --short --porcelain -z", Output: "AM file1\x00"},
+				{Command: "git hash-object -- file1", Output: "hash1\n"},
+				{Command: "git status --short --porcelain -z", Output: "AM file1\x00"},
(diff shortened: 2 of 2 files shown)


scop:
Thanks for looking into this! #1216 is not a well written issue because as mentioned, it actually contains _two_. my bad. This appears to address only one of them, the other "revert+retry when restoring stashed fails" remains as far as I can tell. But nevermind, if/when this is merged, I can file another issue for the remaining parts.

tuchfarber:
> Thanks for looking into this! #1216 is not a well written issue because as mentioned, it actually contains _two_. my bad. This appears to address only one of them, the other "revert+retry when restoring stashed fails" remains as far as I can tell. But nevermind, if/when this is merged, I can file another issue for the remaining parts.

Ah! My bad. My team kept getting bit on this problem on partially staged files so I just trying to fix that. Looking at the issue again, do you have any opinions on comparing the pre-stash changeset or the post-stash changeset? I realize I did post-stage, but your issue sounds like you'd prefer pre-stash.

scop:
Operating on post stash changesets is the right thing to do, i.e.
1. stash unstaged,
2. take changeset snapshot,
3. run tools,
4. take another changeset snapshot,
5. compare changesets taken in 2 and 4, determine if we should fail on changes and output diff,
6. restore stashed changes.

The other issue I'm talking in #1216 is that it's possible that the tools have made changes in step 3) that make the originally stashed changes in step 1) fail to apply when we attempt to restore them in 6). In that case I think the right thing to do would be to remove all changes made by tools in step 3) and attempt restoring again. I.e. what pre-commit does in https://github.com/pre-commit/pre-commit/blob/8a0630ca1aa7f6d5665effe674ebe2022af17919/pre_commit/staged_files_only.py. This is because the changes made by tools can be easily redone, but the stashed changes may contain more valuable work that cannot be automatically/easily redone.

tuchfarber:
That all makes sense! I've updated the `after()` method to take care of that as well. Basically:
1. Deferring the return of the error until after the stash restore is done, and
2. Removing tool changes (`git. HideUnstaged(changes)`) if `RestoreUnstaged()` fails before trying `RestoreUnstaged()` again

p.s. I don't regularly write go code, so holler if I made it too Python-y or anything.

tuchfarber:
@scop Just looking for a ballpark of when this could be reviewed and potentially released. Not trying to rush, just debating if I should move over our laptops to my fork in the meantime to prevent potential data loss.

scop:
Sorry for taking so long getting back to this, and thanks heaps for working on it!

I intend to try and find time to try this out and review soon(tm) (hopefully this week), unless it is merged before that. But I'd like to note that I'm not a lefthook project member but just an individual (opinionated :)) contributor, and it'll be up to @mrexox et al to decide what goes in and how the tool is to behave, so I cannot promise anything regarding merging and releasing. I do hope that if/when I review/approve this, it'll make their job of accepting this a bit easier though ;)

scop:
Added some observations on a quick read through, did not test (will try to do that sometime later).

mrexox:
Hey guys! I'm really sorry for keeping this PR for so long. I am going to merge it and adjust the code a little bit, I'll try to address the original issue, but I need some time. After that I'll release a new version and ask you to take a look and make sure that everything is in the right places.

Thank you for your contribution!

scop:
Thanks!

Simplistic testing with `fail_on_changes: always` looks good to me functionalitywise, nice!

As a user, as a minor improvement, I think I'd prefer to see something like this diagnostic in the output when lefthook forcibly removes changes made by tools to get the stashed ones to apply: "Could not restore stashed changes. reverting ones made by hooks and retrying". IIRC pre-commit and prek do something like this.

mrexox:
I think it makes sense to defined a task for detecting changes and restoring previous state. I still lean to always printing the diff if changes get detected.

If you could prepare a feature-request, and fill it with what's expected, that would be great!

PR #1255, opened 2025-12-22
perf: skip ghost hook when hooks are already configured

### LINKED ISSUE #1254: perf: Avoid redundant prepare-commit-msg ghost hook
### Description
Hi! 👋
I'm managing a large monorepo and noticed that prepare-commit-msg (ghost hook) is always installed even when other hooks are configured.
Since every hook already calls syncHooks() in run.go, the ghost hook becomes redundant when at least one hook exists. In my case, both pre-commit and prepare-commit-msg run on every commit, but only pre-commit does actual work.
This adds ~100-300ms overhead per commit.

```lefthook.yml
pre-commit:
  commands:
    format:
      run: echo "no operation"
```
### Commands to reproduce
```bash
lefthook install
ls .git/hooks/prepare-commit-msg  # exists even though not in config
LEFTHOOK_VERBOSE=1 git commit -m "test"
```

### Lefthook version
```
2.0.12
```

### Possible solution
Only install ghost hook when no hooks are configured:

WooWan:
- Closes #1254 

### Context

As far as I understand, the ghost hook (prepare-commit-msg) is installed to enable seamless synchronization of hooks with the config. However, when there are already configured hooks in the lefthook config, this ghost hook adds unnecessary overhead on every commit.

  Every hook that runs already triggers syncHooks(), which checks if hooks are synchronized with the config and reinstalls them if needed. Therefore, the ghost hook is only needed when there are no configured hooks — it provides a sync mechanism when no other hooks would run.

### Changes

  - Only install ghost hook when there are no existing hooks
  - Updated tests accordingly

### DIFF
--- a/internal/command/install.go
+++ b/internal/command/install.go
@@ -315,7 +315,7 @@ func (l *Lefthook) createHooksIfNeeded(cfg *config.Config, hooks []string, force
 		}
 	}
 
-	if len(onlyHooks) == 0 {
+	if len(onlyHooks) == 0 && len(cfg.Hooks) == 0 {
 		templateArgs := templates.Args{
 			Rc:                      cfg.Rc,
 			AssertLefthookInstalled: cfg.AssertLefthookInstalled,
--- a/internal/command/install_test.go
+++ b/internal/command/install_test.go
@@ -58,8 +58,10 @@ post-commit:
 				configPath,
 				hookPath("pre-commit"),
 				hookPath("post-commit"),
-				hookPath(config.GhostHookName),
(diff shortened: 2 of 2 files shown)


WooWan:
@mrexox  Hello! I'd appreciate it if you could take a look at this PR whenever you have time. Thanks so much!

mrexox:
Ok, it makes sense. Ghost hook was there from the beginning and its task was to make sure you have hooks synced. Then its task was to make sure you don't miss a new hook after editing the config. I think with your change the behavior stays the same. Thank you for putting the effort into this :+1:

PR #1260, opened 2025-12-28
docs: clarify remote settings

### LINKED ISSUE #1247: Unclear documentation for remotes and hooks
### Description

https://lefthook.dev/configuration/remotes.html

> This priority may be changed in the future. For convenience, if you use remotes, please don't configure any hooks.

I think this note deserves some refining. "For convenience" does not seem right, and it's unclear _what_ hooks the recommendation is about, as not defining hooks at all would be kind of pointless I think :)

### Possible solution

I'm not quite sure what is the intent of the note, but taking some guesses, perhaps something like this?

> This priority may be changed in the future. For future proof operation, it is recommended to refrain from configuring any hooks in local main or override configs when using remote configs, i.e. to configure them only in the remote ones.

mrexox:
Closes https://github.com/evilmartians/lefthook/issues/1247

### Context

Clarify the notes on the `remote` option.

### DIFF
--- a/docs/mdbook/configuration/remotes.md
+++ b/docs/mdbook/configuration/remotes.md
@@ -14,4 +14,4 @@ The configuration from `remotes` will be merged to the local config using the fo
 1. Remote configs (`remotes`)
 1. Local overrides (`lefthook-local.yml`)
 
-This priority may be changed in the future. For convenience, try not to have dependencies between your jobs in lefthook.yml config. Remote hooks should be considered as something like a "standalone package of hooks" rather than "plugin with settings extension".
+This priority may be changed in the future. For simplicity, try to keep jobs in remote settings independent from any other steps.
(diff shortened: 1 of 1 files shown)


docs/mdbook/configuration/remotes.md
```diff
@@ -14,4 +14,4 @@ The configuration from `remotes` will be merged to the local config using the fo
 1. Remote configs (`remotes`)
 1. Local overrides (`lefthook-local.yml`)
 
-This priority may be changed in the future. For convenience, if you use `remotes`, please don't configure any hooks.
+This priority may be changed in the future. For convenience, try not to have dependencies between your jobs in lefthook.yml config. Remote hooks should be considered as something like a "standalone package of hooks" rather than "plugin with settings extension".
```
mrexox: @scop WDYT, does it look better?

mrexox: I see. Well, "convenience" is indeed a bad word here, better say robustness and simplicity.

Dependencies between jobs means when you want to have some local jobs with the same names, so that remotes can override them.

And the same thing for "package-like" and "plugin-like" approach. That's actually my intention to provide a best practice to not try having remote jobs for complex and unintuitive flows.

PR #1291, opened 2026-01-21
fix: log full scoped name for skipped jobs

scop:
### Context

When some jobs are skipped, only their "base" name is logged, in contrast with jobs that are not skipped whose full scoped name is logged.

### Changes

Log full scoped name for skipped jobs, too.


mrexox:
Thank you for the contribution! Sorry for a slow review

PR #1339, opened 2026-03-07
fix: separate more commands' non-option args with --

scop:
### Context

Protects e.g. filename arguments (inadvertently?) beginning with a -. Some of these are quite theoretical in nature, but why not.

### Changes

SSIA.

### DIFF
--- a/.lefthook.yml
+++ b/.lefthook.yml
@@ -25,7 +25,7 @@ pre-commit:
 
     - name: check links
       tags: docs
-      run: lychee --max-concurrency 3 {staged_files}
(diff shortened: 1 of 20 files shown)


mrexox:
Looks like integration tests are failing. Could you take a look, please?

scop:
Done, sorry about that :facepalm:

scop:
Hm some more (actually integration ones :facepalm:) remain, for reasons not immediately clear to me.

mrexox:
Oh, it's because these tests use actual configs from the github 😅 . It chould be fine merging the PR!

PR #1353, opened 2026-03-16
fix: skip termenv OSC 11 probe when stdin is not a TTY

dpritchett:
### Context

In git hook contexts stdin is `/dev/null` or a pipe. termenv's OSC 11 background-color probe writes to stdout (a TTY) but reads the response from the same fd. The terminal sends its reply to the input side, where nothing consumes it, so the raw escape sequence leaks to the screen:

```
┃  snag-filter ❯

^[]11;rgb:2020/2020/2020^[\^[[49;1R
```

`NO_COLOR=1` does not help -- that suppresses color output, not terminal probing. `TERM=dumb` works because termenv skips probing entirely.

This is a known termenv bug (https://github.com/muesli/termenv/issues/189) affecting multiple consumers including `gh` CLI. A termenv-level fix would make this workaround deletable, but lefthook should defend itself regardless.

### Changes

`configureRenderer` in `NewLefthook()` checks whether stdin is a TTY. When it is not, it pins the default lipgloss renderer to `termenv.ANSI` with a dark background assumption, which prevents the OSC probe from firing. Colors still render through `CompleteColor` ANSI fallbacks.

This lives in `NewLefthook()` rather than `log.New()` or an `init()` because `NewLefthook()` already owns color configuration and runs once per invocation. Keeping it here avoids import-time side effects and keeps the function testable (it takes `*os.File` so tests can pass a pipe).

### DIFF
--- a/internal/command/lefthook.go
+++ b/internal/command/lefthook.go
@@ -9,11 +9,14 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/charmbracelet/lipgloss"
 	"github.com/knadh/koanf/parsers/json"
 	"github.com/knadh/koanf/providers/rawbytes"
 	"github.com/knadh/koanf/v2"
+	"github.com/muesli/termenv"
 	"github.com/spf13/afero"
 	"github.com/urfave/cli/v3"
+	"golang.org/x/term"
 
 	"github.com/evilmartians/lefthook/v2/internal/config"
 	"github.com/evilmartians/lefthook/v2/internal/git"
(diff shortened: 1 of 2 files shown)


dpritchett:
Closing this — the OSC 11 probe is triggered by snag (a hook command), not lefthook itself. Lefthook doesn't use `AdaptiveColor` or call `HasDarkBackground()`, so the probe never fires from lefthook's own code.

The interaction is subtler than I initially thought: lefthook runs hook commands inside a child pty (`pty.Start()`), which makes snag's TTY guard pass when it shouldn't. The actual fix belongs in snag (https://github.com/dpritchett/snag/issues/46).

Upstream root cause: https://github.com/muesli/termenv/issues/205

Sorry for the noise.

PR #1383, opened 2026-04-10
fix(windows): normalize lefthook path for sh script

AndrewKahr:
### Context
On Windows, the paths generated to the lefthook binary contain backslashes, which the sh script interprets as escape characters. This causes all calls to lefthook to fail with `Can't find lefthook in PATH`.

Note: This seems to be masked by yarn, which is able to pick it up. However, NPM alone is unable to do so without the path normalization.

### Changes
This PR updates `templates.go` by calling `filepath.ToSlash` to normalize the slashes in `LefthookPath` and `LefthookPathCurrent`. Additionally, it adds an integration test to ensure Lefthook is successfully installed and detected in a basic capacity in an NPM project. Finally, all main workflow actions have been bumped to resolve deprecation issues.


mrexox:
Thank you for the PR! I'll review it later next week when I am able to run my Windows machine :+1:

AndrewKahr:
Just to make sure this edge case gets testing in the future, I added a test workflow to verify that NPM successfully detects the path on Windows (and for good measure, Mac and Linux runners too). If I run this workflow against the master branch, it errors as expected, and on the updated branch, I can verify it passes. I also bumped all the main workflow actions to their current versions to resolve the deprecation notices from the upcoming node version update on hosted Github runners.

internal/templates/templates.go
```diff
@@ -46,8 +47,8 @@ func Hook(hookName string, args Args) []byte {
 		Rc:                      args.Rc,
 		AssertLefthookInstalled: args.AssertLefthookInstalled,
 		Roots:                   args.Roots,
-		LefthookPath:            strings.ReplaceAll(strings.TrimSpace(args.LefthookPath), "\n", ";"),
-		LefthookPathCurrent:     lefthookPathCurrent,
+		LefthookPath:            filepath.ToSlash(strings.ReplaceAll(strings.TrimSpace(args.LefthookPath), "\n", ";")),
+		LefthookPathCurrent:     filepath.ToSlash(lefthookPathCurrent),
```
mrexox, line 51: This change is good, but let's remove the changes in test.yml workflow. This code can be tested in unit tests

AndrewKahr: Sounds good, I've removed the new job from test.yml

mrexox:
Awesome! Thank you for preparing the PR!

AndrewKahr:
Glad I could help! Looking forward to the next update release

PR #1386, opened 2026-04-12
fix: normalize root option to always include trailing slash

### LINKED ISSUE #1380: `root:` without trailing slash causes double slashes in resolved file paths
### Description

When using the `root:` configuration in job option **without a trailing slash** (e.g., `root: "client"` instead of `root: "client/"`), lefthook produces file paths with double slashes like `.//src/index.js` instead of `./src/index.js`.

The `root` option is documented as a **directory** — it changes the CWD for command execution and filters file paths by directory.
Users understand `root` as "the directory where my command runs," not as "a string prefix to match against file paths."
No user would expect that `root: "client"` and `root: "client/"` behave differently, because both unambiguously refer to the same directory.

However, the current implementation treats `root` as a raw string prefix for replacement rather than as a directory path, which causes the double slash.

The [documentation](https://lefthook.dev/configuration/root/) shows an example with a trailing slash (`root: "client/"`), but does not state that a trailing slash is required. The JSON schema also defines `root` as a plain string with no constraints.

### `lefthook.yml`

```yaml
pre-commit:
  commands:
    lint-no-slash:
      root: "client"
      glob: "*.{js,ts}"
      run: echo "root without slash:" {staged_files}
    lint-with-slash:
      root: "client/"
      glob: "*.{js,ts}"
      run: echo "root with slash:" {staged_files}
```

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true

# Setup repro in /tmp
rm -rf /tmp/lefthook-repro
mkdir -p /tmp/lefthook-repro/client/src
cd /tmp/lefthook-repro

git init
git commit --allow-empty -m "init"

echo '{}' > client/package.json
echo 'console.log("hello")' > client/src/index.js

cat > lefthook.yml << 'CONF'
pre-commit:
  commands:
    lint-no-slash:
      root: "client"
      glob: "*.{js,ts}"
      run: echo "root without slash:" {staged_files}
    lint-with-slash:
      root: "client/"
      glob: "*.{js,ts}"
      run: echo "root with slash:" {staged_files}
CONF

git add .
git commit -m "initial"

echo "modified" >> client/src/index.js
git add client/src/index.js

lefthook run pre-commit
```

Output:
```
sync hooks: ✔️ (pre-commit)
│  [lefthook] git: git status --short --porcelain -z
│             out: M  client/src/index.js
│  [lefthook] git: git diff --name-only --cached --diff-filter=ACMRD
│             out: client/src/index.js
│  [lefthook] git: git diff --name-only --cached --diff-filter=ACMR
│             out: client/src/index.js
│  [lefthook] filtered [ ]: client/src/index.js
│             filtered [x]: .//src/index.js
│  [lefthook] files after escaping: .//src/index.js
│  [lefthook] job: echo "root without slash:" .//src/index.js
│  [lefthook] run: echo "root without slash:" .//src/index.js
┃  lint-no-slash ❯

root without slash: .//src/index.js

│  [lefthook] filtered [ ]: client/src/index.js
│             filtered [x]: ./src/index.js
│  [lefthook] files after escaping: ./src/index.js
│  [lefthook] job: echo "root with slash:" ./src/index.js
│  [lefthook] run: echo "root with slash:" ./src/index.js
┃  lint-with-slash ❯

root with slash: ./src/index.js
```

**Expected:** Both commands resolve to `./src/index.js`.
**Actual:** `root: "client"` (no trailing slash) resolves to `.//src/index.js` (double slash).

### Lefthook version

```
2.1.5
```

### Possible solution

The issue is in [`internal/run/controller/filter/filter.go`, line 193](https://github.com/evilmartians/lefthook/blob/master/internal/run/controller/filter/filter.go#L193), in the `byRoot()` function:

```go
func byRoot(vs []string, matcher string) []string {
        if matcher == "" {
                return vs
        }

        vsf := make([]string, 0)
        for _, v := range vs {
                if strings.HasPrefix(v, matcher) {
                        vsf = append(vsf, strings.Replace(v, matcher, "./", 1))
                }
        }
        return vsf
}
```

The function replaces the `root` prefix with `"./"` via naive string replacement. When `matcher` is `"client"` (no trailing slash), `"client/src/index.js"` becomes `".//src/index.js"` because the `/` separator is preserved.

Normalize `matcher` to always have a trailing slash before performing the replacement:

```go
func byRoot(vs []string, matcher string) []string {
        if matcher == "" {
                return vs
        }

        if !strings.HasSuffix(matcher, "/") {
                matcher += "/"
        }

        vsf := make([]string, 0)
        for _, v := range vs {
                if strings.HasPrefix(v, matcher) {
                        vsf = append(vsf, strings.Replace(v, matcher, "./", 1))
                }
        }
        return vsf
}
```

The test expectations in [`filter_test.go`, lines 198-207](https://github.com/evilmartians/lefthook/blob/master/internal/run/controller/filter/filter_test.go#L198) also encode the double-slash as expected and would need to be updated.

Yanhu007:
Fixes #1380

## Problem

When `root` is set without a trailing slash (e.g. `root: "client"`), `byRoot` produces paths with double slashes:

```
root: "client"  → .//src/index.js  (wrong)
root: "client/" → ./src/index.js   (correct)
```

The `strings.Replace(v, matcher, "./", 1)` replaces `"client"` with `"./"` but leaves the `/` separator after it.

## Fix

Normalize the matcher to always end with `"/"` before filtering. Also switch from `strings.Replace` to `strings.TrimPrefix` for clarity.

Updated existing tests to reflect the corrected behavior (no more double slashes).

### DIFF
--- a/internal/run/controller/filter/filter.go
+++ b/internal/run/controller/filter/filter.go
@@ -187,10 +187,16 @@ func byRoot(vs []string, matcher string) []string {
 		return vs
 	}
 
+	// Ensure matcher ends with a path separator so that
+	// root: "client" and root: "client/" behave identically.
+	if !strings.HasSuffix(matcher, "/") {
+		matcher += "/"
+	}
+
 	vsf := make([]string, 0)
 	for _, v := range vs {
 		if strings.HasPrefix(v, matcher) {
-			vsf = append(vsf, strings.Replace(v, matcher, "./", 1))
+			vsf = append(vsf, "./"+strings.TrimPrefix(v, matcher))
 		}
 	}
 	return vsf
--- a/internal/run/controller/filter/filter_test.go
+++ b/internal/run/controller/filter/filter_test.go
@@ -198,12 +198,12 @@ func TestByRoot(t *testing.T) {
 		{
 			source: []string{"folder/subfolder/0.rb", "subfolder/1.txt", "folder/2.RB", "3.rbs"},
 			path:   "folder",
-			result: []string{".//subfolder/0.rb", ".//2.RB"},
+			result: []string{"./subfolder/0.rb", "./2.RB"},
 		},
 		{
 			source: []string{"folder/subfolder/0.rb", "folder/1.rbs"},
 			path:   "folder/subfolder",
-			result: []string{".//0.rb"},
+			result: []string{"./0.rb"},
 		},
 		{
 			source: []string{"folder/subfolder/0.rb", "folder/1.rbs"},


mrexox:
Closing in favor of https://github.com/evilmartians/lefthook/pull/1381

PR #1393, opened 2026-04-14
fix: skip pty allocation when stdout is not a terminal

### LINKED ISSUE #1392: All `run:` commands fail with EPERM in sandboxed environments (pty.Start calls setsid)
### Summary

Every `run:` command fails instantly with "operation not permitted" when lefthook runs inside a macOS sandbox that restricts `setsid()`. This includes trivial commands like `echo hello`. Git commands work fine.

The root cause is `pty.Start()` in `exec_unix.go`, which sets `SysProcAttr.Setsid = true` via `creack/pty`. The sandbox allows pty device access (`/dev/ptmx`, `/dev/ttys*`) but blocks creating a new session. Since the pty path is the default for all non-interactive commands, there's no config-level workaround that preserves the intended semantics.

Related: #1354 mentions this in the context of AI agent detection, but the pty issue is independent of AI agents and affects any sandboxed environment.

### Lefthook version

```
2.1.4
```

### Steps to reproduce

1. Run lefthook inside any macOS sandbox that restricts `setsid()` (Claude Code, OpenAI Codex, or a custom `sandbox-exec` profile)
2. `lefthook run pre-push --force`

### Expected results

Commands execute normally.

### Actual results

Every `run:` command fails instantly:

```
operation not permitted
operation not permitted
operation not permitted

summary: (done in 0.21 seconds)
🥊 rust-fmt (0.00 seconds)
🥊 rust-clippy (0.00 seconds)
🥊 go-fmt (0.12 seconds)
🥊 shell-lint (0.02 seconds)
🥊 ruby-syntax (0.02 seconds)
🥊 bundle-install (0.00 seconds)
🥊 check-links (0.02 seconds)
🥊 bats (0.00 seconds)
```

Git commands within lefthook (version check, diff, status) all succeed because they go through plain `exec.Command.Run()` with no `SysProcAttr`.

### Root cause

In `exec_unix.go`, the default execution path (line 93-101) calls `pty.Start(command)`. Under the hood, `creack/pty` does:

```go
cmd.SysProcAttr.Setsid = true
cmd.SysProcAttr.Setctty = true
```

The `setsid()` syscall creates a new session, which sandboxes block. The error happens before the child process starts, so the actual command is irrelevant.

The `interactive: true` path (line 85-92) uses plain `command.Start()` with no `SysProcAttr`, which works fine.

### Workarounds tested

| Option | Bypasses pty? | Result |
|--------|--------------|--------|
| `interactive: true` | **Yes** | Works, but semantic mismatch (commands aren't interactive) |
| `follow: true` | No | Still fails |
| `--no-tty` | No | Still fails |
| `--no-colors` | No | Still fails |

### Suggestion

Skip pty allocation when stdout isn't a terminal:

```go
if args.interactive || args.useStdin || !isatty.IsTerminal(os.Stdout.Fd()) {
    // direct exec path
} else {
    // pty path (current behavior)
}
```

`isatty` is already imported.

For color preservation on the non-pty path, lefthook already injects `CLICOLOR_FORCE=true` when `colors` is explicitly `"on"` (exec_unix.go line 54). The same injection could apply when taking the non-pty path, so tools that respect [`CLICOLOR_FORCE`](http://bixense.com/clicolors/) (cargo, cmake, and others) still produce colored output.

This would fix sandboxed environments, CI runners, IDE commit UIs (VS Code, GitHub Desktop), and piped output without needing to detect specific environments.

technicalpickles:
## Summary

`pty.Start()` fails with EPERM in sandboxed environments (Claude Code, Codex, etc.) because the sandbox blocks access to `/dev/ptmx` and the associated ioctl calls needed for pseudo-terminal allocation. This breaks all `run:` commands.

The pty exists to make child processes see a terminal on stdout so they emit colored output. When stdout is already not a terminal (sandbox, CI, pipes), the pty adds nothing useful.

**Changes:**
- Check `isatty.IsTerminal(os.Stdout.Fd())` before taking the pty path
- When stdout is not a tty, use direct exec instead
- Merge stderr into the stdout buffer on the non-pty path to match pty behavior (where both streams go through the same device)

No new dependencies. `isatty` is already imported.

## Behavior matrix

| Environment | Before | After |
|---|---|---|
| Normal terminal | pty (colors work) | pty (unchanged) |
| Sandbox | **EPERM, all commands fail** | direct exec (commands work) |
| CI (piped stdout) | pty (colors via pty trick) | direct exec (use `colors: on` for colors) |
| Pipes (`\| tee log`) | pty (ANSI in piped output) | direct exec (clean piped output) |

## Reproduction

Self-contained repro script: https://gist.github.com/technicalpickles/f24ba1cc62764a50b9074a1bc518831d

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Repro script: 7/7 scenarios pass in sandbox (was 3/7)
- [x] Repro script: 7/7 scenarios pass outside sandbox
- [x] Stderr routing verified: output matches pty baseline (merged in formatted blocks, parallel jobs cleanly separated)
- [ ] Manual testing in real terminal with colored linter output

Fixes #1392

### DIFF
--- a/internal/run/controller/exec/exec_unix.go
+++ b/internal/run/controller/exec/exec_unix.go
@@ -90,7 +90,7 @@ func (e CommandExecutor) execute(ctx context.Context, cmdstr string, args *execu
 		if err != nil {
 			return err
 		}
-	} else {
+	} else if isatty.IsTerminal(os.Stdout.Fd()) {
 		p, err := pty.Start(command)
 		if err != nil {
 			return err
@@ -99,6 +99,17 @@ func (e CommandExecutor) execute(ctx context.Context, cmdstr string, args *execu
 		defer func() { _ = p.Close() }()
 
 		_, _ = io.Copy(args.out, p)
+	} else {
+		// No pty available (sandbox, CI, pipe). Merge stderr into
+		// stdout buffer to match pty behavior where both streams
+		// go through the same device.
+		command.Stdout = args.out
+		command.Stderr = args.out
+		command.Stdin = args.in
+		err := command.Start()
+		if err != nil {
+			return err
+		}
 	}
 
 	defer func() { _ = command.Process.Kill() }()
--- a/internal/run/controller/exec/exec_unix.go
+++ b/internal/run/controller/exec/exec_unix.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"syscall"
 
 	"github.com/creack/pty"
 	"github.com/mattn/go-isatty"
@@ -82,7 +83,8 @@ func (e CommandExecutor) execute(ctx context.Context, cmdstr string, args *execu
 	command.Dir = args.root
 	command.Env = append(os.Environ(), args.envs...)
 
-	if args.interactive || args.useStdin {
+	switch {
+	case args.interactive || args.useStdin:
 		command.Stdout = args.out
 		command.Stdin = args.in
 		command.Stderr = os.Stderr
@@ -90,7 +92,7 @@ func (e CommandExecutor) execute(ctx context.Context, cmdstr string, args *execu
 		if err != nil {
 			return err
 		}
-	} else if isatty.IsTerminal(os.Stdout.Fd()) {
+	case isatty.IsTerminal(os.Stdout.Fd()):
 		p, err := pty.Start(command)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


technicalpickles:
I started https://github.com/evilmartians/lefthook/pull/1394 to add test coverage for this code path, which should hopefully fix the codecov difference complaining.

technicalpickles:
Built this branch locally and tested it in a macOS sandbox (Claude Code). Two validations:

**Repro script (7 scenarios, 7 pass)**

Ran [the reproduction gist](https://gist.github.com/technicalpickles/f24ba1cc62764a50b9074a1bc518831d) from #1392 against this branch. All 7 scenarios pass, including the ones that previously failed with EPERM (default, follow, no_tty, interactive+no_tty).

Every scenario correctly takes the direct exec path (`stdout is a tty: no`) instead of the pty path.

**Real project (13 hooks, all pass)**

Also ran `lefthook run pre-push --all-files` on a project with a pretty beefy `lefthook.yml` (cargo fmt, cargo clippy, gofmt, shellcheck, ruby syntax, rspec, bats, lychee, buildkite-builder). All 13 commands passed in ~20 seconds, zero EPERM failures. Previously every single one of these failed instantly.

technicalpickles:
Added the full test suite from #1394 plus a concurrent output isolation test for the default (non-pty) path.

Two variants of the isolation test:

- `TestExecute_ConcurrentOutputIsolation_Interactive`: uses `Interactive: true` for the direct exec path. Works on master, establishes the contract.
- `TestExecute_ConcurrentOutputIsolation`: uses the default exec path. Only passes with this fix, since on master it would hit `pty.Start` → `setsid` → EPERM in any sandboxed/CI environment where stdout isn't a terminal.

Both spin up 5 goroutines writing 20 labeled lines each with sleeps, verify each buffer only contains its own worker's output. Mirrors how the controller runs parallel jobs.

mrexox (approved):

PR #1394, opened 2026-04-14
test(exec): add unit tests for unix command executor

technicalpickles:
## Summary

- Adds first test coverage for the `internal/run/controller/exec` package, which previously had zero test files
- Tests both the default execution path (pty on terminal, direct exec otherwise) and the interactive/useStdin path
- Covers: output capture, stderr merging, env propagation, working directory, error handling, multi-command sequencing, context cancellation, and stdin passthrough

## Context

Extracted from #1393 to land test coverage independently. These tests exercise the existing behavior so that the upcoming non-pty fallback change has a regression safety net.

## Test plan

- [x] `go test -v ./internal/run/controller/exec/` passes (10/10)
- [x] `make lint` clean (0 issues)
- [x] `make test` passes (all existing packages unaffected)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/run/controller/exec/exec_unix_test.go
+++ b/internal/run/controller/exec/exec_unix_test.go
@@ -1,0 +1,134 @@
+//go:build !windows
+
+package exec
+
+import (
+	"bytes"
+	"context"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExecute(t *testing.T) {
+	tmpDir := t.TempDir()
+	// Resolve symlinks so pwd -P output matches (macOS: /var -> /private/var).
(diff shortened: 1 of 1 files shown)


technicalpickles:
Added a concurrent output isolation test (`TestExecute_ConcurrentOutputIsolation_Interactive`). It spins up 5 goroutines that each run a command writing 20 labeled lines with sleeps, all sharing nothing but the executor. Each goroutine gets its own `bytes.Buffer`, same as the controller does for parallel jobs. Asserts that each buffer only contains its own worker's output and nothing leaked from the others.

Uses `Interactive: true` to take the direct exec path, so it works on master without any changes to the executor.

technicalpickles:
All tests from this PR were included in #1393 (now merged as part of v2.1.6), including the concurrent output isolation test added in the latest push. Nothing unique remains on this branch, so closing.

PR #1396, opened 2026-04-17
fix: separate fallback push branch from pathspecs

### LINKED ISSUE #1395: "fatal: ambiguous argument" when a folder exists that matches the branch name
### Description
If a repo has a folder that happens to have the same name a branch, the default left hook pre push hook will fail. 

E.g.  for a repo with a branch name `dev` and a folder named `dev`:

```
│  > git diff --name-only HEAD dev                              
│    fatal: ambiguous argument 'dev': both revision and filename
│    Use '--' to separate paths from revisions, like this:      
│    'git <command> [<revision>...] -- [<file>...]'
```

### Lefthook version

2.1.4 

### Possible solution

Claude's analysis: 

internal/git/repository.go

Roughly:

Constants near the top of that file — cmdPushFilesHead is the git diff --name-only HEAD prefix:
`cmdPushFilesHead = []string{"git", "diff", "--name-only", "HEAD"}`
PushFiles() — first it tries HEAD vs @{push} (cmdPushFilesBase). If that fails, it resolves a short branch name into r.headBranch and then does the problematic append:
`return r.FindExistingFiles(append(cmdPushFilesHead, r.headBranch), "")`
That becomes a shell/git invocation equivalent to:

`git diff --name-only HEAD <short-branch-name>`

When `<short-branch-name>` is dev and your worktree also has a path dev/, Git errors: both revision and filename.

lawrence3699:
Closes #1395

### Context

When `@{push}` is unavailable, `PushFiles()` falls back to `git diff --name-only HEAD <default-branch>`.
If the repository also contains a directory with the same name as that default branch, Git treats the branch name as ambiguous and the pre-push hook fails before it can resolve `{push_files}`.

### Changes

- terminate the fallback `git diff` revision list with `--`
- add regression coverage for the branch/path collision case

### Validation

- `go test ./internal/git -run TestPushFiles -count=1`
- `go test ./internal/git -count=1`
- `go test ./internal/run/controller/... -count=1`
- `go test ./... -count=1`
- manual repro on a throwaway repo where `origin/HEAD` points to `dev` and the worktree contains `dev/`

### DIFF
--- a/internal/git/repository.go
+++ b/internal/git/repository.go
@@ -194,7 +194,11 @@ func (r *Repository) PushFiles() ([]string, error) {
 		return r.FindExistingFiles(cmdLsTreeFilesHead, "")
 	}
 
-	return r.FindExistingFiles(append(cmdPushFilesHead, r.headBranch), "")
+	command := make([]string, 0, len(cmdPushFilesHead)+2)
+	command = append(command, cmdPushFilesHead...)
+	command = append(command, r.headBranch, "--")
+
+	return r.FindExistingFiles(command, "")
 }
 
 // resolveHeadBranch determines the upstream head branch.
--- a/internal/git/repository_test.go
+++ b/internal/git/repository_test.go
@@ -223,4 +223,59 @@ func TestPushFiles(t *testing.T) {
 			t.Fatalf("expected %v, got %v", want, files)
 		}
 	})
+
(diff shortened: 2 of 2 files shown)


mrexox (approved):

mrexox:
Thank you for preparing the PR! I'll fix linter issues

PR #1434, opened 2026-05-30
refactor: use services for commands

mrexox:
Closes # (issue)

### Changes

<h3>Confidence Score: 1/5</h3>

Not safe to merge — the build does not compile and two core commands are no-ops.

Three distinct compile errors will prevent any binary from being produced. Even if those were patched, `lefthook run` and `lefthook uninstall` both return nil without doing anything, silently breaking the two most common user workflows. Additionally, `loadConfig` dereferences a potentially nil pointer and `ForEachRemote` can crash the process via panic on bad config.

`internal/commands/run.go`, `internal/commands/uninstall.go`, and `internal/app/hooks_service.go` have compile errors. `internal/app/config_service.go` needs attention for the nil-deref, panic, and unsafe type assertions.

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `internal/app/config_service.go`, line 625-633 ([link](https://github.com/evilmartians/lefthook/blob/515597f9d9037bfa04503275c099183a2433fb89/internal/app/config_service.go#L625-L633)) 

   <a href="#">P1</a> **Unchecked type assertions in `setColors` can panic**

   In the `map[string]any` branch, each color is extracted with `colorsTyped["cyan"].(string)` and similar. If a key is absent from the map the assertion is on `nil`, and if the value is not actually a string both cases produce a runtime panic. The config is user-supplied, so a missing or wrong-typed color name is entirely plausible. Use the two-value assertion form (`v, ok := colorsTyped["cyan"].(string)`) and skip or fall back gracefully when `!ok`.

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: rebase the branch \[skip ci\]"](https://github.com/evilmartians/lefthook/commit/515597f9d9037bfa04503275c099183a2433fb89) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42707081)</sub>

> Greptile also left **8 inline comments** on this PR.


mrexox:
@greptileai review

PR #1443, opened 2026-06-12
deps: migrate shellescape to al.essio.dev/pkg/shellescape

maguowei:
Title: deps: migrate shellescape to al.essio.dev/pkg/shellescape

---

## Why

The `shellescape` module was renamed from `github.com/alessio/shellescape` to `al.essio.dev/pkg/shellescape` starting with v1.5.0 (see https://github.com/alessio/shellescape/releases).

Lefthook itself builds fine because its graph pins the old path at v1.4.1. However, downstream projects that pull lefthook in as a Go tool dependency and resolve a fresh module graph can end up selecting a post-rename version of the old path, which fails with:

```
github.com/alessio/shellescape@v1.6.0: parsing go.mod:
        module declares its path as: al.essio.dev/pkg/shellescape
                but was required as: github.com/alessio/shellescape
```

Other projects hit the same issue with this dependency, e.g. kubernetes-sigs/kind#3840.

## What

- Switch the three import sites (`build_command.go`, `build_script.go`, `replacer/replacer.go`) to the new canonical path
- Bump the dependency to v1.6.0 in `go.mod` / `go.sum`

Package name is unchanged (`shellescape`), so no code changes are needed. `go mod tidy`, `go build ./...` and `go test ./...` pass locally.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 )
 
 require (
-	github.com/alessio/shellescape v1.4.1
+	al.essio.dev/pkg/shellescape v1.6.0
 	github.com/fatih/color v1.18.0 // indirect
 	github.com/lucasb-eyer/go-colorful v1.4.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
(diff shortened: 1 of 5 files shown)


mrexox (approved):
Thank you for taking care!

PR #1448, opened 2026-06-18
feat: AI coding agents integration

### LINKED ISSUE #1433: Add LLM hooks management support
### Description

Claude and Codex have hooks, and managing them through lefthook feels like something useful. 

### What problem it is solving?

You don't need to define hooks for the Claude, Codex, etc. You add a hook to `lefthook.yml` and generate the configs for the provider of your choice.

### Details

Need to think of the better approach and configuration option.

#### Option 1 – Strictly configurable

```yml
# lefthook.yml

ai:
  claude:
    Stop: validate # references a named lefthook hook

validate:
  jobs:
    - run: go test ./...
```

Merges to:

```jsonc
// .claude/settings.json

{
  ... // other local user settings
  "hooks": {
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "lefthook run validate"
          }
        ]
      }
    ]
  }
} 
```

#### Option 2 – Ask user

```yml
Stop:
  jobs:
    - run: go test ./...
```

When you run `lefthook install`

1. Detects you have configured hooks for LLM (by name)
2. Asks if you prefer to use Claude or Codex
  2.1. Another option is to check existing .claude/ and .codex/ dirs
4. Installs the hook (which call `lefthook run Stop`)

viralpraxis:
resolves https://github.com/evilmartians/lefthook/issues/1433

some decisions/proposals:
  - we're following the first path ("Strictly configurable"). I think that since provider-specific event sets are not equal, it'd be better to delegate configuration to user. I also don't think it's worth letting `lefthook` know bout specific events (we'll have to synchronize them from time to time0;

TODO:
  - [x] make sure that cleanup round-trips with `ai` configuration option
  - [x] revisit autogenerated tests and make sure they are sane


internal/command/install_ai.go
```diff
@@ -0,0 +1,278 @@
...
+func validateAIHooks(ai *config.AI, hooks map[string]*config.Hook) error {
+	var missing []string
+
+	check := func(provider string, events map[string]string) {
+		for event, hookName := range events {
+			if _, ok := hooks[hookName]; !ok {
+				missing = append(missing, fmt.Sprintf("ai.%s.%s -> %q", provider, event, hookName))
+			}
+		}
+	}
```
mrexox: Let's move this to a separated `func checkAIHooks`

internal/command/install_ai.go
```diff
@@ -0,0 +1,278 @@
...
+
+	check("claude", ai.Claude)
+	check("codex", ai.Codex)
+
+	if len(missing) > 0 {
+		slices.Sort(missing)
+		return fmt.Errorf("ai config references undefined hooks: %s", strings.Join(missing, ", "))
```
mrexox: I think it makes sense to log error with one error on each line, then return `ai hooks misconfigured` error. Just for readability

internal/command/install_ai.go
```diff
@@ -0,0 +1,278 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// removeAIHookEntries reads a provider settings file, strips lefthook-managed
+// entries, and writes the result back. If nothing remains, the file is removed.
+func (l *Lefthook) removeAIHookEntries(path string) error {
```
mrexox: I think these functions could be moved to `uninstall_ai.go` just for code separation.

internal/command/install_ai.go
```diff
@@ -0,0 +1,278 @@
...
+	return true
+}
+
+// buildHookEntry constructs a single matcher group that runs `lefthook run <hookName>`.
+func buildHookEntry(hookName string) map[string]any {
+	return map[string]any{
+		"hooks": []any{
+			map[string]any{
+				"type":    "command",
+				"command": lefthookRunPrefix + hookName,
```
mrexox: As a first iteration I think it's acceptable. But most lefthook users use NPM package and don't have lefthook installed globally. I think we could use the result of `os.Executable()` to cover the case of locally installed lefthook, WDYT?

mrexox (commented):
LGTM! A few nitpicks, but the overall approach looks good :+1:

internal/command/install_ai.go
viralpraxis: Makes sense, tried it out in https://github.com/evilmartians/lefthook/pull/1448/changes/bf1205262d084532362de02d749622692c3a07f9

viralpraxis: On the other hand, if users decide not to VCS-ignore project-scoped hooks, using bare `lefhook` would be better -- in that case they'll have to set `lefthook` configuration option to `"lefthook"`

PR #1453, opened 2026-06-24
docs: fix language

pflanze:
(There is no pre-existing issue.)

### Context

I've been reading the documentation and stumbled reading the docs index page in two places. Here are my suggestions for improving it.

### Changes

These seem stylistic issues, I hope that people agree that my changes are more fluid to read / better standard English sentences.


mrexox:
Thank you so much! Indeed better language brings better understanding!

PR #1464, opened 2026-07-08
fix: preserve quoting of run commands and forwarded script args

### LINKED ISSUE #1167: sh syntax error when running on windows
### Description

sh syntax error when running on windows: `-c: line 3: syntax error: unexpected end of file` .

### `lefthook.yml`

```yaml
colors: false
test:
  jobs:
    - name: test
      run: |
        if [ -z "$CI" ]; then
            echo "NO CI"
        else
            echo "CI"
        fi
```

### Commands to reproduce

```powershell
$ $env:LEFTHOOK_VERBOSE='true'

$ lefthook run test
│  [lefthook] git: git version
│             out: git version 2.51.0.windows.2
│  [lefthook] git: git rev-parse --path-format=absolute --show-toplevel --git-path hooks --git-path info --git-dir
│             out: C:/Users/xxxx/workspaces/projects/xxx
│                  C:/Users/xxxx/workspaces/projects/xxx/.git/hooks
│                  C:/Users/xxxx/workspaces/projects/xxx/.git/info
│                  C:/Users/xxxx/workspaces/projects/xxx/.git
│  loading config: C:\Users\xxxx\workspaces\projects\xxx\lefthook.yaml
╭──────────────────────────────╮
│ lefthook v1.13.6  hook: test │
╰──────────────────────────────╯
│  [lefthook] run: "C:\Program Files\Git\bin\sh.exe" -c "if [ -z "$CI" ]; then
│      echo "NO CI"
│  else
│      echo "CI"
│  fi
│  "
⠙ waiting: testCI
else
    echo CI
fi
: -c: line 3: syntax error: unexpected end of file
│  [lefthook] git: git diff --name-only --cached --diff-filter=ACMR
│  [lefthook] git: git diff --name-only --cached --diff-filter=ACMRD
│  [lefthook] git: git status --short --porcelain
│             out: M lefthook-nodejs.yaml
│                  M lefthook.yaml
│                  ?? CHANGELOG.md
┃  test ❯



summary: (done in 0.19 seconds)
✗ test (0.19 seconds)
```

### Lefthook version

`1.13.6 539f66c92f10e20ed369d769afee1cd6e93d5735`

### Possible solution

Thank you for creating this excellent software!

I found the problem occurs here:

https://github.com/evilmartians/lefthook/blob/57a23f95e78ce154fbd16fbc6173b04e79dd1d80/internal/run/controller/exec/exec_windows.go#L114

I'm not familiar with the golang and don't understand why it can't be passed as a parameter like unix:

https://github.com/evilmartians/lefthook/blob/57a23f95e78ce154fbd16fbc6173b04e79dd1d80/internal/run/controller/exec/exec_unix.go#L82

Or we can also use a temp file or stdin for sh to read to avoid this problem.

Bineroflux:
Closes #1167

### Context

Two related quoting bugs mangle command strings on their way to `sh`:

1. **Forwarded args lose their quoting (all platforms).** `lefthook run <hook> -- <args...>` joins the args into the script command string without escaping: args with spaces arrive split, and `lefthook run my-hook -- "file (1).txt"` fails with ``sh: syntax error near unexpected token `(` ``.

2. **The Windows executor mangles any command containing a double quote (#1167).** It hand-builds the raw process command line as `"sh" -c "cmdstr"` without escaping, so the msys argument parser corrupts the command before `sh` sees it:

| command string | today on Windows | with this fix (= Linux/macOS) |
|---|---|---|
| `echo "hello world"` | `hello` | `hello world` |
| `printf '<%s>' "$VAR"` | splits on spaces | verbatim |
| multi-line `run:` with quotes (#1167) | syntax error | works |

Since even `echo "hello world"` mis-executes today, no config can rely on the old behavior — the fix restores parity with other platforms and matches how the `skip`/`only` check executor already invokes `sh`.

### Changes

- `build_script.go` — quote each forwarded arg with `shellescape.Quote`, like the script path and files templates already are. Args without special characters stay untouched, so `-- -Apply`-style flag forwarding is unchanged.
- `exec_windows.go` — drop the hand-built `SysProcAttr.CmdLine`; pass `-c cmdstr` as regular arguments and let `os/exec` quote it, same as the unix executor. Also resolves the "consider quoting" comment in that file.
- New unit test for built script commands, plus two integration tests (`run_script_forwarded_args`, `run_sh_quotes`) that fail on `master` and pass with this change.

Verified on Windows 11 (Git for Windows) and Ubuntu: `go test ./...`, the full integration suite, and end-to-end runs with `sh` and `powershell -File` scripts receiving filenames with spaces, parentheses, and quotes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Confidence Score: 5/5</h3>

Both changes are minimal, targeted, and verified by new tests; the Windows path now matches the Unix executor exactly.

The forwarded-arg fix is a straightforward one-liner wrapped in a loop, confined to the already-quoted else-branch. The Windows executor change removes a hand-rolled quoting scheme that was provably broken and replaces it with the same exec.CommandContext call the Unix executor has used all along. Both directions of the fix are covered by a new unit test and two integration tests that reproduce the original bugs. No behaviour is changed on paths that already worked.

No files require special attention.

<sub>Reviews (1): Last reviewed commit: ["fix: pass the command to sh verbatim on ..."](https://github.com/evilmartians/lefthook/commit/042e48326f8f95e7ca82adaf870eb81dd1880886) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42707052)</sub>

### DIFF
--- a/internal/run/controller/command/build_script.go
+++ b/internal/run/controller/command/build_script.go
@@ -85,7 +85,11 @@ func (b *Builder) buildScript(params *JobParams) ([]string, []string, error) {
 			commands, _ := replacer.ReplaceAndSplit(command, system.MaxCmdLen())
 			execs = append(execs, commands...)
 		} else {
-			args = append(args, b.opts.GitArgs...)
+			// Quote the args to pass them to the script as-is, even if they contain
+			// spaces or shell special characters.
+			for _, gitArg := range b.opts.GitArgs {
+				args = append(args, shellescape.Quote(gitArg))
+			}
 			execs = append(execs, strings.Join(args, " "))
 		}
 	}
--- a/internal/run/controller/command/build_script_test.go
+++ b/internal/run/controller/command/build_script_test.go
@@ -1,0 +1,79 @@
+package command
+
+import (
+	"path/filepath"
+	"testing"
+
+	"al.essio.dev/pkg/shellescape"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/evilmartians/lefthook/v2/tests/helpers/gittest"
+	"github.com/evilmartians/lefthook/v2/tests/helpers/loggertest"
+)
+
+func TestBuildScript(t *testing.T) {
+	root, err := filepath.Abs("src")
+	assert.NoError(t, err)
+
+	hookName := "echo-args"
+	sourceDir := filepath.Join(root, ".lefthook")
+	scriptPath := filepath.Join(sourceDir, hookName, "echo.sh")
+	quotedScript := shellescape.Quote(scriptPath)
+
+	for name, tt := range map[string]struct {
+		runner  string
+		gitArgs []string
+		exec    string
+	}{
+		"no args": {
+			runner: "sh",
+			exec:   "sh " + quotedScript,
+		},
+		"simple args are passed as-is": {
+			runner:  "sh",
+			gitArgs: []string{"-Apply", "file.txt"},
+			exec:    "sh " + quotedScript + " -Apply file.txt",
+		},
+		"args with spaces are kept intact": {
+			runner:  "powershell",
+			gitArgs: []string{"path with spaces.yml", "file (1).txt"},
+			exec:    "powershell " + quotedScript + " 'path with spaces.yml' 'file (1).txt'",
+		},
+		"args with shell special characters are escaped": {
+			runner:  "sh",
+			gitArgs: []string{"a$b", "c`d", "e&f;g", "h\"i"},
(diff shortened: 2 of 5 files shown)


mrexox:
@greptileai review, please

Bineroflux:
@mrexox Friendly reminder. Are you interested in this PR? Or should I close it?

mrexox:
@Bineroflux thank you for reminding. I have quite busy time now, so wasn't able to review and check the changes (you could notice the growing number of issues and PRs). But I'll catch up! This PR looks good, I just need to do some testing on a Windows machine. Thank you for taking care!

PR #1472, opened 2026-07-27
docs: quote template run examples

### LINKED ISSUE #1467: Documented unquoted template syntax causes YAML parsing error
### Description

Using a template at the beginning of an unquoted `run` value causes Lefthook to fail while parsing the configuration:

```text
Error: yaml: line 8: did not find expected key
```

The current [`templates` documentation](https://lefthook.dev/configuration/templates/) recommends this syntax:

```yaml
- run: {wrapper} yarn format
```

However, `{` begins a flow mapping in YAML. It therefore cannot begin an unquoted plain scalar. Lefthook fails during YAML parsing before template substitution occurs.

Quoting the value or using a block scalar resolves the error:

```yaml
- run: "{wrapper} yarn format"
```

```yaml
- run: |
    {wrapper} yarn format
```

### `lefthook.yml`

```yaml
templates:
  wrapper: npm exec --no --

pre-commit:
  jobs:
    - name: Prettier
      run: {wrapper} prettier --check .
```

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true
npx lefthook validate
```

Alternatively:

```bash
export LEFTHOOK_VERBOSE=true
npx lefthook run pre-commit --all-files
```

Observed result:

```text
Error: yaml: line 6: did not find expected key
```

Expected result: the configuration is parsed and `{wrapper}` is replaced with its configured value, as shown in the documentation.

### Lefthook version

```text
2.1.10 8d9cfec5f52367af6374a5430b4e9568844856e5
```

### Possible solution

Update the documentation examples to quote `run` values that begin with a template:

```yaml
- run: "{wrapper} yarn format"
- run: "{wrapper} yarn lint"
- run: "{wrapper} yarn test"
```

The earlier `{dip}` example on the same page should be quoted as well:

```yaml
- run: "{dip} bundle exec rubocop -- {staged_files}"
```

Alternatively, Lefthook could preprocess leading template placeholders before YAML parsing, though documenting YAML-valid quoted values is likely simpler and less surprising.

Guflly:
Closes #1467

### Context

YAML treats a leading `{` in an unquoted value as the start of a flow mapping, so the documented commands fail to parse before template substitution.

### Changes

Quote the four `run` examples that begin with `{dip}` or `{wrapper}`.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge because the documentation examples now parse correctly without altering the commands users intend to run.

YAML removes the delimiting quotes when decoding these values, so the examples retain the intended template commands while avoiding the invalid unquoted leading-brace form.

<sub>Reviews (1): Last reviewed commit: ["docs: quote template run examples"](https://github.com/evilmartians/lefthook/commit/61d3e421500168324e75e69c0173ebd6d4ac6ce1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47692583)</sub>

### DIFF
--- a/docs/configuration/templates.md
+++ b/docs/configuration/templates.md
@@ -23,7 +23,7 @@ templates:
 pre-commit:
   jobs:
     # Will run: `bundle exec rubocop -- file1 file2 file3 ...`
-    - run: {dip} bundle exec rubocop -- {staged_files}
+    - run: "{dip} bundle exec rubocop -- {staged_files}"
 ```
 
 ```yml
@@ -43,7 +43,7 @@ templates:
 
 pre-commit:
   jobs:
-    - run: {wrapper} yarn format
-    - run: {wrapper} yarn lint
-    - run: {wrapper} yarn test
+    - run: "{wrapper} yarn format"
+    - run: "{wrapper} yarn lint"
+    - run: "{wrapper} yarn test"
 ```


mrexox (approved):
Thank you!

PR #1509, opened 2026-08-28
fix: quote paths in the generated hook shim

### LINKED ISSUE #1488: Generated hook shim interpolates paths unquoted: a space in the path silently disables all hooks
### Description

`internal/templates/hook.tmpl` interpolates three path variables **unquoted**. When any of them contains a space, the generated hook word-splits it, every branch of `call_lefthook()` fails, and the hook silently no-ops — printing `Can't find lefthook in PATH` while the commit **still succeeds with exit 0**.

The failing branch is not stale or missing. The binary exists, is executable, and runs fine when the same path is quoted. It is purely a quoting bug.

The three unquoted interpolations, all in `internal/templates/hook.tmpl` (still present at `v2.1.10`):

```gotemplate
[ -f {{.Rc}} ] && . {{.Rc}}                          {{/* .Rc — unquoted twice */}}

elif test -n "{{ .LefthookPath }}"                   {{/* test IS quoted... */}}
then
  {{ .LefthookPath }} "$@"                           {{/* ...but the call is not */}}

elif {{ .LefthookPathCurrent }} -h >/dev/null 2>&1   {{/* both unquoted */}}
then
  {{ .LefthookPathCurrent }} "$@"
```

Note the inconsistency inside this same file: every `$dir/node_modules/...` candidate below is correctly quoted (`"$dir/node_modules/lefthook/bin/index.js"`), so the fix matches the surrounding style.

This appears to be a regression of #403 ("Space in path cause lefthook to fail", closed COMPLETED in 2022), which reported the identical symptom at v1.2.6.

### Why it is worse than it looks

Because the terminal branch `echo "Can't find lefthook in PATH"` exits 0 (absent `assert_lefthook_installed`), the result is not a visible failure — it is **silent, permanent skipping of every hook**. Formatting and guards never run locally and the problem only surfaces in CI, if at all.

On macOS this is easy to hit without doing anything unusual: `~/Library/Mobile Documents/...` (iCloud Drive), `~/Google Drive/My Drive/...`, and any project folder with a space in its name.

### `lefthook.yml`

```yaml
pre-commit:
  commands:
    hello:
      run: echo HOOK_ACTUALLY_RAN
```

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true

# a path containing a space, and a lefthook binary living inside it
mkdir -p "/tmp/my project/bin"
cp "$(which lefthook)" "/tmp/my project/bin/lefthook"
cd "/tmp/my project"
git init -q . && git config user.email t@t.t && git config user.name t
printf 'pre-commit:\n  commands:\n    hello:\n      run: echo HOOK_ACTUALLY_RAN\n' > lefthook.yml

# install USING that binary, so its path is baked into the hook
"/tmp/my project/bin/lefthook" install

# the generated shim contains the unquoted path:
grep -n "my project" .git/hooks/pre-commit

echo x > f.txt && git add f.txt && git commit -m test
```

**Actual:**

```
Can't find lefthook in PATH
[main (root-commit) eb1b24a] test
```

The `hello` command never runs. The commit succeeds.

The generated lines (verbatim):

```sh
  elif /tmp/my project/bin/lefthook -h >/dev/null 2>&1
  then
    /tmp/my project/bin/lefthook "$@"
```

`sh` splits that into `/tmp/my` + `project/bin/lefthook`, so the `elif` can never be true.

**Control — identical setup, no space in the path:**

```
┃  hello ❯
HOOK_ACTUALLY_RAN
✔️ hello (0.02 seconds)
```

The space is the only variable.

### Lefthook version

Reproduced empirically on `2.1.9 75f99ff325a731c296b572292dd1c855cd970b93` (macOS 26.5.2, arm64, git 2.50.1, `/bin/sh`).

I did not rebuild `v2.1.10` to re-run the repro, but I read `internal/templates/hook.tmpl` at tag `v2.1.10` and all three interpolations above are still unquoted there, so I expect it to reproduce unchanged on latest.

### Possible solution

Quote the three interpolations:

```gotemplate
[ -f "{{.Rc}}" ] && . "{{.Rc}}"

elif test -n "{{ .LefthookPath }}"
then
  "{{ .LefthookPath }}" "$@"

elif "{{ .LefthookPathCurrent }}" -h >/dev/null 2>&1
then
  "{{ .LefthookPathCurrent }}" "$@"
```

That is sufficient for spaces. A double quote in a path would still break it, so escaping the value at render time would be more robust if you want to close the general case.

### Relationship to #1398

#1398 reports the same `LefthookPathCurrent` branch failing, but attributes it to the baked absolute path going **stale** (pointing into a deleted git worktree) and proposes deleting the branch outright.

This report is a distinct and independent defect in the same lines: the branch fails **even when the baked path is perfectly valid and the target exists**, purely from word-splitting. That means

- it reproduces in a plain single-checkout repo with no worktrees involved (as above), and
- for any repo under a path containing a space, that branch has never worked at all and the shim has always been silently relying on the `$dir/...` fallbacks.

The two are complementary: quoting fixes the general case, and #1398's proposal would independently remove one of the three affected interpolations. Happy to reference whichever direction you prefer.

addielaruee:
Closes #1488

### Context

When `lefthook install` generates a hook, it bakes the path to the running binary (`os.Executable()`, the `.LefthookPathCurrent` template value) into the shim. That path was interpolated unquoted, so a binary living under a directory with a space (common on macOS: iCloud Drive, Google Drive, any folder with a space in its name) is word-split by the generated `sh`, every branch of `call_lefthook()` fails, and the shim prints `Can't find lefthook in PATH`. Because that terminal branch exits `0` (unless `AssertLefthookInstalled` is set), the commit still succeeds, so the hook is silently and permanently skipped with no signal. Unlike the config values, this path is detected at install time, so the user cannot quote it themselves.

Reproduced with the issue's steps (a lefthook binary living under `/tmp/my project/bin`):

- Before: the shim contains `elif /tmp/my project/bin/lefthook -h ...`; committing prints `Can't find lefthook in PATH` and succeeds with the hook never running.
- After: the shim contains `elif '/tmp/my project/bin/lefthook' -h ...`; committing runs the hook.

### Changes

- Added a `shellescape` template function (POSIX single quotes, `'\''` escaping) and applied it to the two `.LefthookPathCurrent` interpolations in `hook.tmpl`.
- Left the `lefthook` and `rc` config values unquoted on purpose: `lefthook` is documented as an executable path or command (e.g. `bundle exec lefthook`, `LEFTHOOK_VERBOSE=1 lefthook`), and `rc` is documented with tilde and environment-variable expansion (e.g. `~/.lefthookrc`, `"${XDG_CONFIG_HOME:-$HOME/.config}/lefthookrc"`). Those rely on the shell interpreting them, and the docs make quoting a spaced path there the user's responsibility. Escaping them would turn commands into nonexistent filenames and stop documented expansions from resolving.
- Added `internal/templates/templates_test.go`: table-driven tests that the auto-detected path is single-quoted, and that a configured command, an env-prefixed command, and tilde/env-expansion rc paths are passed through unquoted.

Verified locally: `go test ./internal/...` and `golangci-lint run` (pinned v2.11.4) pass. No config structs changed, so `schema.json` is untouched.

### DIFF
--- a/internal/templates/templates.go
+++ b/internal/templates/templates.go
@@ -40,8 +40,10 @@ func Hook(hookName string, args Args) []byte {
 	}
 
 	buf := &bytes.Buffer{}
-	t := template.Must(template.ParseFS(templatesFS, "hook.tmpl"))
-	if err = t.Execute(buf, hookTmplData{
+	t := template.Must(template.New("hook.tmpl").Funcs(template.FuncMap{
+		"shellescape": shellescape,
+	}).ParseFS(templatesFS, "hook.tmpl"))
+	if err = t.ExecuteTemplate(buf, "hook.tmpl", hookTmplData{
 		HookName:                hookName,
 		Extension:               getExtension(),
 		Rc:                      args.Rc,
@@ -74,4 +76,14 @@ func getExtension() string {
 		return ".exe"
 	}
 	return ""
+}
+
+// shellescape wraps a value in POSIX single quotes so it is passed to the
+// generated hook verbatim. It is used for the auto-detected executable path
+// (os.Executable), which the user cannot quote themselves: a path with a
+// space was word-split and silently disabled every hook. The user-supplied
+// `lefthook` and `rc` values are intentionally not escaped, since they are
+// documented as commands and shell-expanded paths.
+func shellescape(value string) string {
+	return "'" + strings.ReplaceAll(value, "'", `'\''`) + "'"
 }
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -33,9 +33,9 @@ call_lefthook()
     lefthook.bat "$@"
   {{ end -}}
   {{ if .LefthookPathCurrent -}}
-  elif {{ .LefthookPathCurrent }} -h >/dev/null 2>&1
+  elif {{ shellescape .LefthookPathCurrent }} -h >/dev/null 2>&1
   then
-    {{ .LefthookPathCurrent }} "$@"
+    {{ shellescape .LefthookPathCurrent }} "$@"
   {{ end -}}
   else
     dir="$(git rev-parse --show-toplevel)"
--- a/internal/templates/templates_test.go
+++ b/internal/templates/templates_test.go
@@ -1,0 +1,85 @@
+package templates
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func renderHook(t *testing.T, data hookTmplData) string {
+	t.Helper()
+	buf := &bytes.Buffer{}
+	tmpl := template.Must(template.New("hook.tmpl").Funcs(template.FuncMap{
+		"shellescape": shellescape,
+	}).ParseFS(templatesFS, "hook.tmpl"))
+	require.NoError(t, tmpl.ExecuteTemplate(buf, "hook.tmpl", data))
+	return buf.String()
+}
+
+func TestShellescape(t *testing.T) {
+	for name, tt := range map[string]struct {
+		value string
+		want  string
+	}{
+		"plain path":            {value: "/usr/bin/lefthook", want: `'/usr/bin/lefthook'`},
+		"path with space":       {value: "/home/my user/bin/lefthook", want: `'/home/my user/bin/lefthook'`},
+		"embedded single quote": {value: "/home/o'brien/lefthook", want: `'/home/o'\''brien/lefthook'`},
+		"metacharacters":        {value: "/tmp/$(touch pwned)", want: `'/tmp/$(touch pwned)'`},
+		"empty":                 {value: "", want: `''`},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert.Equal(t, tt.want, shellescape(tt.value))
+		})
+	}
+}
+
+func TestHookQuotesAutodetectedExecutablePath(t *testing.T) {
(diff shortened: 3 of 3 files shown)


addielaruee:
Good catch. The first revision single-quoted all three interpolated values, which was wrong: `lefthook` is documented as an executable path or command (e.g. `bundle exec lefthook`) and `rc` supports tilde and `$VAR` expansion (e.g. `~/.lefthookrc`), so quoting those breaks documented behavior.

I narrowed the fix to only `.LefthookPathCurrent`, the path detected from `os.Executable()` at install time. That value is always a real filesystem path and the user has no way to quote it themselves, so escaping it is safe and is the actual case in #1488. The `lefthook` and `rc` values are now left untouched, and the tests assert that a configured command and tilde/env-expansion rc paths pass through unquoted.

internal/templates/templates.go
```diff
@@ -75,3 +77,13 @@ func getExtension() string {
...
 }
+
+// shellescape wraps a value in POSIX single quotes so it is passed to the
+// generated hook verbatim. It is used for the auto-detected executable path
+// (os.Executable), which the user cannot quote themselves: a path with a
+// space was word-split and silently disabled every hook. The user-supplied
+// `lefthook` and `rc` values are intentionally not escaped, since they are
+// documented as commands and shell-expanded paths.
+func shellescape(value string) string {
+	return "'" + strings.ReplaceAll(value, "'", `'\''`) + "'"
```
mrexox, line 88: Hey @addielaruee ! There's a dedicated `shellescape` package in the depdendencies. Could we use it instead of manual replacements?

PR #1519, opened 2026-08-31
fix(exec): expose lefthook binary on PATH for nested hook commands

### LINKED ISSUE #1518: `lefthook` not found when called with `push` hook command
### Description

When invoked from another hook (using `uv run lefthook run ...`), the following `lefthook-validate` command works fine. But when running `git push`, `PATH` appears to be reset (perhaps by Git itself).

### `lefthook.yml`

```yaml
# ...
    lefthook-validate:
      run: 'lefthook validate'
      use_stdin: false
```

### Commands to reproduce

```sh
export LEFTHOOK_VERBOSE=true
```

```text
...
┃  lefthook-validate ❯ 
sh: lefthook: command not found
...
```

### Lefthook version

`2.1.10 8d9cfec5f52367af6374a5430b4e9568844856e5`

### Possible solutions

Alternatives:

- Provide the path to the currently running `lefthook` process as an env var.
- Provide a configuration flag, similar to `min_version`, to enforce a validation before `run`.
- Override Lefthook and/or Git's behavior to not reset `PATH` or at least Lefthook's path. This seems to have a lot of intricacies and security implications, I suppose.

rwinkelman:
## Summary
- When git invokes hooks (especially `pre-push`), the environment often has a stripped `PATH`
- Nested jobs like `run: lefthook validate` then fail with `sh: lefthook: command not found` even though the parent process is lefthook
- Prepend the running binary's directory to `PATH` and set `LEFTHOOK_BIN` for every job command

Fixes #1518

## Test plan
- [x] Unit tests for `withLefthookEnv` (`env_test.go`)
- [ ] CI `go test ./internal/run/controller/exec/...`

### DIFF
--- a/internal/run/controller/exec/env.go
+++ b/internal/run/controller/exec/env.go
@@ -1,0 +1,65 @@
+package exec
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const lefthookBinEnv = "LEFTHOOK_BIN"
+
+// withLefthookEnv prepends the running lefthook binary's directory to PATH and
+// sets LEFTHOOK_BIN so nested hook commands like `run: lefthook validate` work
+// when git invokes hooks with a stripped PATH (see issue #1518).
+func withLefthookEnv(env []string) []string {
+	exe, err := os.Executable()
+	if err != nil || exe == "" {
+		return env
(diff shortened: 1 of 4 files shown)


rwinkelman:

PR #1520, opened 2026-08-31
fix(run): propagate job exit codes for agent hooks

### LINKED ISSUE #1511: ai: Claude Code Stop/PreToolUse hooks need exit code 2 to block, but `lefthook run` always exits 1
## Summary

The `ai:` integration (2.1.10, #1448) generates `.claude/settings.json` entries that run `lefthook run <hook>`. Claude Code only treats a hook as *blocking* when the command exits with code **2**: for `Stop`, exit 2 blocks the stop and feeds stderr back to the agent so it keeps working; for `PreToolUse`, exit 2 blocks the tool call. Any other non-zero exit is "non-blocking": stderr is shown to the user and the agent stops / the tool call proceeds.

`lefthook run` exits **1** whenever a job fails, even when the job itself exits 2. So a `Stop: validate` hook that fails lint just prints to the user after the agent has already finished; the agent never sees it.

## Reproduction

```yaml
# lefthook.yml
ai:
  claude:
    Stop: validate
validate:
  jobs:
    - run: exit 2
```

```
$ lefthook install
$ lefthook run validate; echo $?
...
exit status 2
...
1
```

Same with `run: exit 1` → lefthook exits 1. Claude Code docs on exit codes: https://code.claude.com/docs/en/hooks#hook-output

## Suggestion

When invoked from a generated agent hook (or via a flag, e.g. `lefthook run --agent validate` that the `ai:` generator emits), exit 2 on failure — or propagate the highest job exit code. Also worth noting in the docs that Claude passes `stop_hook_active: true` on stdin when the agent is already continuing from a Stop hook; users who lint on Stop will want to exit 0 in that case to avoid a loop.

lefthook 2.1.10, macOS (Homebrew).

rwinkelman:
## Summary
- Claude Code Stop/PreToolUse hooks only block when the command exits **2**; `lefthook run` previously always exited **1** on failure
- When `LEFTHOOK_AGENT=1`, propagate the highest failing job exit code to the process exit status
- Generated `ai:` hook commands now prefix `LEFTHOOK_AGENT=1` automatically

Fixes #1511

## Test plan
- [x] Updated `install_ai_test.go` expectations for generated commands
- [ ] CI unit/integration tests

### DIFF
--- a/internal/command/exitcode.go
+++ b/internal/command/exitcode.go
@@ -1,0 +1,12 @@
+package command
+
+// HookExitError is returned from `lefthook run` when hooks fail. Main uses
+// Code as the process exit status so agent integrations (Claude Code Stop /
+// PreToolUse) can block on exit code 2.
+type HookExitError struct {
+	Code int
+}
+
+func (e *HookExitError) Error() string {
+	return ""
+}
--- a/internal/command/install_ai.go
+++ b/internal/command/install_ai.go
@@ -60,7 +60,7 @@ func lefthookRunCommand(bin, hookName string, quoteBin bool) string {
 		bin = shellQuotePath(bin)
 	}
 
-	return bin + lefthookRunSuffix + hookName
+	return "LEFTHOOK_AGENT=1 " + bin + lefthookRunSuffix + hookName
 }
 
 func shellQuotePath(path string) string {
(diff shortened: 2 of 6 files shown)


rwinkelman:

PR #1521, opened 2026-08-31
fix(templates): quote hook shim paths that contain spaces

### LINKED ISSUE #1488: Generated hook shim interpolates paths unquoted: a space in the path silently disables all hooks

rwinkelman:
## Summary
- Quote `Rc`, `LefthookPath`, and `LefthookPathCurrent` in generated git hook shims so paths containing spaces do not word-split and silently skip all hooks.

Fixes #1488

## Test plan
- [x] Added `internal/templates/templates_test.go` covering spaced paths and shell escaping
- [ ] CI

### DIFF
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -10,7 +10,7 @@ fi
 
 {{- if .Rc}}
 {{/* Load rc file, which may export ENV variables */}}
-[ -f "{{.Rc}}" ] && . "{{.Rc}}"
+[ -f {{.Rc}} ] && . {{.Rc}}
 {{- end}}
 
 call_lefthook()
@@ -18,10 +18,10 @@ call_lefthook()
   if test -n "$LEFTHOOK_BIN"
   then
     "$LEFTHOOK_BIN" "$@"
-  {{ if .LefthookPath -}}
-  elif test -n "{{ .LefthookPath }}"
+  {{ if .LefthookPathInvoke -}}
+  elif test -n {{ .LefthookPathTest }}
   then
-    "{{ .LefthookPath }}" "$@"
+    {{ .LefthookPathInvoke }} "$@"
   {{ end -}}
   elif lefthook{{.Extension}} -h >/dev/null 2>&1
   then
@@ -33,9 +33,9 @@ call_lefthook()
     lefthook.bat "$@"
   {{ end -}}
   {{ if .LefthookPathCurrent -}}
-  elif "{{ .LefthookPathCurrent }}" -h >/dev/null 2>&1
+  elif {{ .LefthookPathCurrentTest }} -h >/dev/null 2>&1
   then
-    "{{ .LefthookPathCurrent }}" "$@"
+    {{ .LefthookPathCurrent }} "$@"
   {{ end -}}
   else
     dir="$(git rev-parse --show-toplevel)"
--- a/internal/templates/hook.tmpl
+++ b/internal/templates/hook.tmpl
@@ -10,7 +10,7 @@ fi
 
 {{- if .Rc}}
 {{/* Load rc file, which may export ENV variables */}}
-[ -f {{.Rc}} ] && . {{.Rc}}
+[ -f "{{.Rc}}" ] && . "{{.Rc}}"
 {{- end}}
 
 call_lefthook()
@@ -21,7 +21,7 @@ call_lefthook()
   {{ if .LefthookPath -}}
   elif test -n "{{ .LefthookPath }}"
   then
-    {{ .LefthookPath }} "$@"
+    "{{ .LefthookPath }}" "$@"
   {{ end -}}
   elif lefthook{{.Extension}} -h >/dev/null 2>&1
   then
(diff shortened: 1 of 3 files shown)


rwinkelman:
Pushed a follow-up addressing Greptile feedback:

- Quote path-like `lefthook` config values only when they contain spaces (absolute/relative paths), not arbitrary shell commands like `bundle exec lefthook`.
- Stop escaping `$` in rc paths so values like `${XDG_CONFIG_HOME:-$HOME/.config}/lefthookrc` still expand when sourced.
- Split test vs invoke quoting in the hook template so `test -n` stays correct for both paths and commands.

Ready for another review pass when CI finishes.

rwinkelman:

PR #1522, opened 2026-08-31
fix(install): honor lefthook-local.yml without creating lefthook.yml

### LINKED ISSUE #1489: `lefthook install` should recognize `lefthook-local.yml` without creating `lefthook.yml`
### Description

According to [the documentation](https://lefthook.dev/configuration/), `lefthook-local.yml` can be used without a main configuration file:

> The `-local` config can be used without a main config file.

However, when only `lefthook-local.yml` exists, `lefthook install` reports that no config was found and creates a sample `lefthook.yml`.

Setting `LEFTHOOK_CONFIG=lefthook-local.yml` is a possible workaround, but `lefthook install` should recognize the documented local-only configuration natively.

### `lefthook.yml`

none

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true
```

1. Create a repository without `lefthook.yml`.
2. Create only `lefthook-local.yml`:
    ```yaml
    pre-commit:
        jobs:
            - name: check
              run: echo check
    ```
3. Run `lefthook install`
4. Lefthook creates `lefthook.yml` and install hooks

### Lefthook version

2.1.10

### Possible solution

If the project has `lefthook-local.yml`, `lefthook install` should install hooks without `lefthook.yml`

rwinkelman:
## Summary
- `lefthook install` no longer scaffolds `lefthook.yml` when only `lefthook-local.yml` (or other local config names) exists, matching documented behavior.
- Checksum timestamp resolution uses the same config discovery path.

Fixes #1489

## Test plan
- [x] Added install test for local-only config
- [ ] CI

### DIFF
--- a/internal/command/install.go
+++ b/internal/command/install.go
@@ -115,8 +115,29 @@ func (l *Lefthook) readOrCreateConfig() (*config.Config, error) {
 }
 
 func (l *Lefthook) configExists(path string) bool {
-	configPath, _ := l.findMainConfig(path)
-	return configPath != ""
+	_, err := l.findConfig(path)
+	return err == nil
+}
+
+func (l *Lefthook) findConfig(path string) (string, error) {
+	if configPath, err := l.findMainConfig(path); err == nil {
+		return configPath, nil
+	}
+
+	return l.findLocalConfig(path)
+}
+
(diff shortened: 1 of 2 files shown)


rwinkelman:

PR #1527, opened 2026-09-01
test(cmd): regression for valid fish shell completion output

### LINKED ISSUE #1430: Fish shell completion file contains invalid syntax
### Description

When using **fish shell**, typing `lefthook<space>` after a fresh **Homebrew** install triggers the following error:

```
lefthook /opt/homebrew/share/fish/vendor_completions.d/lefthook.fish (line 3): Unsupported use of '='. In fish, please use 'set string lefthook'.
function __%!_(string=lefthook)perform_completion
               ^~~~~~~~~~~~~~^
from sourcing file /opt/homebrew/share/fish/vendor_completions.d/lefthook.fish
source: Error while reading file '/opt/homebrew/share/fish/vendor_completions.d/lefthook.fish'
```

This appears to be caused by an upstream bug in urfave/cli, which lefthook uses for CLI scaffolding. See https://github.com/urfave/cli/issues/2285

### Lefthook version

2.1.8

### Possible solution

Looks like lefthook already uses version v3.9.0 of urfave/cli (which contains the fix), so I'm not sure what needs to happen to fix this.

rwinkelman:
## Summary
- Adds a regression test for `lefthook completion fish` so we never ship fish scripts containing invalid Go default-value syntax like `(string=lefthook)`.
- Documents the fix already present via `urfave/cli/v3` v3.10.1 (see urfave/cli#2285).

Fixes #1430 (verification/regression coverage).

## Test plan
- [x] `go test ./cmd -run TestFishCompletionValidSyntax`

### DIFF
--- a/cmd/completion_test.go
+++ b/cmd/completion_test.go
@@ -1,0 +1,34 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+)
+
+func TestFishCompletionValidSyntax(t *testing.T) {
+	root := Lefthook()
+	var buf bytes.Buffer
+	root.Writer = &buf
+
+	if err := root.Run(context.Background(), []string{"lefthook", "completion", "fish"}); err != nil {
+		t.Fatalf("completion fish: %v", err)
+	}
+
+	out := buf.String()
+	if out == "" {
+		t.Fatal("expected fish completion script")
+	}
+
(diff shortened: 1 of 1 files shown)


rwinkelman:

rwinkelman:
Rebased onto current `master` (v2.1.14); restored test commit after accidental empty push. Grooming pass.

mrexox:
@greptileai review

PR #1536, opened 2026-09-09
docs: document files inheritance for grouped jobs

### LINKED ISSUE #1528: Document files behavior for grouped jobs
### Description

The documentation for [`group`](https://lefthook.dev/configuration/group/) describes several job options that are inherited by nested jobs, but it does not appear to document the behavior of `files`.

With Lefthook 2.1.12, a `files` command defined on a parent job containing a `group` is evaluated for the nested jobs, and `{files}` in each nested job receives the resulting file list.

This behavior appears intentional and works correctly, but it is not clearly documented.

### `lefthook.yml`

```yaml
pre-commit:
  jobs:
    - name: Print Python files
      glob:
        - "*.py"
      files: |
        printf '%s\n' test.py README.md
      group:
        jobs:
          - name: First
            run: |
              printf 'first: %s\n' {files}

          - name: Second
            run: |
              printf 'second: %s\n' {files}
```

With this configuration, both nested jobs receive `test.py` after the parent `glob` is applied.

A practical example is:

```yaml
pre-commit:
  jobs:
    - name: Ruff
      glob:
        - "*.py"
      files: |
        if git diff --cached --quiet -- .ruff.toml; then
          git diff --cached --name-only --diff-filter=ACMR
        else
          git ls-files
        fi
      group:
        jobs:
          - name: Lint
            run: |
              uv run --no-sync ruff check --fix {files}

          - name: Format
            run: |
              uv run --no-sync ruff format {files}
```

Here, the custom `files` command is evaluated for both nested Ruff jobs.

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true

touch test.py README.md
git add test.py README.md

lefthook run pre-commit
```

The verbose output shows the parent `files` command being evaluated for each nested job and `{files}` resolving to `test.py`.

### Lefthook version

```text
2.1.12
```

### Possible solution

Document the behavior of `files` on grouped jobs in the `group` documentation, alongside the other parent-level options that apply to nested jobs.

It would be useful to state explicitly whether `files` is inherited or otherwise evaluated in the parent scope and made available to nested jobs through `{files}`.

chiliec:
Closes #1528

### Context

The `group` docs list `env`, `root`, `glob` and `exclude` as options inherited by nested jobs, but not `files`. In practice a `files` command set on a group is inherited too (`scope.extend` carries `filesCmd` down), so `{files}` in each nested job resolves to the parent command's output, and a nested job can set its own `files` to override it.

### Changes

- Add `files` to the list of inherited group options and note that it is evaluated for each nested job that uses `{files}` and can be overridden per job.
- Extend the existing example with `files` on the group and an override on the second nested job.

Docs only, no code changes.

Checked against a local build of `master` (Go 1.26.6): a `files` command on the group is executed for each nested job and `{files}` resolves to its output; a nested job with its own `files` uses that instead. The updated example passes `lefthook validate` / `lefthook dump`.

Happy to adjust the wording.

### DIFF
--- a/docs/configuration/group.md
+++ b/docs/configuration/group.md
@@ -25,7 +25,7 @@ pre-commit:
           - run: echo 3
 ```
 
-If you specify `env`, `root`, `glob`, or `exclude` on a group, they will be inherited to the underlying jobs.
+If you specify `env`, `root`, `glob`, `exclude`, or `files` on a group, they will be inherited to the underlying jobs. The `files` command is executed for every nested job that uses the `{files}` template; a nested job can define its own `files` to override it.
 
 ```yml
 # lefthook.yml
@@ -39,13 +39,15 @@ pre-commit:
       exclude:
         - "README.md"
       root: "subdir/"
+      files: git diff --name-only master
       group:
         parallel: true
         jobs:
-          - run: echo $E1
-          - run: echo $E1
+          - run: echo $E1 {files}
+          - run: echo $E1 {files}
             env:
               E1: bonjour
+            files: git ls-files
 ```
 
 ::: callout info Note
--- a/docs/configuration/group.md
+++ b/docs/configuration/group.md
@@ -25,7 +25,7 @@ pre-commit:
           - run: echo 3
 ```
 
-If you specify `env`, `root`, `glob`, `exclude`, or `files` on a group, they will be inherited to the underlying jobs. The `files` command is executed for every nested job that uses the `{files}` template; a nested job can define its own `files` to override it.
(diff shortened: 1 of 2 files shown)


chiliec:
Good catches, both addressed:

- Moved the `files` behaviour into its own example without `root`, so the nested jobs actually receive files (checked with a local build: the inherited `git diff --name-only master` gives the first job the changed `.py` file, the overriding `git ls-files` gives the second job all tracked `.py` files).
- Updated `jobs.md` to list `env` and `files` alongside `root`/`glob`/`exclude` so the two pages agree.

mrexox (approved):
Thank you for taking care!

PR #1544, opened 2026-09-14
fix: don't write an install-time lefthook path into AI hook configs

### LINKED ISSUE #1482: ai hooks write an absolute binary path into committed .claude/settings.json
### Description

With `ai.claude` configured, `lefthook install` writes an absolute path to the lefthook binary into `.claude/settings.json`:

```json
"command": "/Users/me/project/node_modules/.pnpm/lefthook-darwin-arm64@2.1.10/node_modules/lefthook-darwin-arm64/bin/lefthook run validate"
```

`.claude/settings.json` is normally committed and shared with the team. That path contains the home directory, the OS and arch (`darwin-arm64`), the lefthook version, and if you use git worktrees, the worktree it was generated in. So the generated config only works for the person who ran install. It also breaks on Linux CI, and after any lefthook upgrade.

#1433, which specified this feature, shows the intended output as:

```json
"command": "lefthook run validate"
```

Two smaller things in the same output. The generated entries have no `matcher`, so a `PostToolUse` mapping runs on every tool call instead of only on file edits. And there is no way to set a `timeout`; Claude Code defaults to 60 seconds, which a full typecheck plus test run exceeds.

Related: #1398 reports the same problem in the `.git/hooks` shim.

### `lefthook.yml`

```yml
ai:
  claude:
    Stop: validate

validate:
  jobs:
    - run: echo hi
```

### Commands to reproduce

```bash
export LEFTHOOK_VERBOSE=true

mkdir repro && cd repro && git init
npm install lefthook
# add the lefthook.yml above
npx lefthook install
cat .claude/settings.json
```

### Lefthook version

```
2.1.10 8d9cfec5f52367af6374a5430b4e9568844856e5
```

Installed via pnpm (`lefthook-darwin-arm64@2.1.10`), macOS arm64.

### Possible solution

Resolve the binary at runtime instead of baking in the path, the way the git hook template already does with its `call_lefthook` function. For npm and pnpm installs, `npx lefthook run <hook>` or `pnpm exec lefthook run <hook>` is portable on its own.

Alternatively, an option to write `.claude/settings.local.json` instead. Claude Code reads that file and it is gitignored by convention, so a machine specific path in it does no harm.

```yml
ai:
  claude:
    local: true       # write .claude/settings.local.json
    Stop: validate
```

anandghegde:
Closes #1482

### Context

`lefthook install` with an `ai:` section writes the result of `os.Executable()` into the generated hook commands. With the npm package that ends up as something like `/Users/me/project/node_modules/.pnpm/lefthook-darwin-arm64@2.1.10/.../bin/lefthook run validate` in `.claude/settings.json`. That file is normally committed, so the command only works for whoever ran install, and breaks on other machines, on CI and after upgrades.

### Changes

- When the `lefthook` config option is set it is still used as is.
- Otherwise the generated command is plain `lefthook` (per review). Projects that run lefthook from a package manager can set `lefthook: npx lefthook` / `pnpm exec lefthook`.
- Paths are no longer embedded, so the shell-quoting helper is gone.
- Entries written by earlier versions still contain `lefthook run `, so re-running install replaces them. Added a test for that.
- Updated the note in `docs/configuration/ai.md`.

This is the AI-hook counterpart of #1398. #1526 (open) removes the `os.Executable()` branch from the git hook shim; this PR touches different files and doesn't depend on it.

Verified:
- `go test ./...` passes, `golangci-lint run` reports no new issues.
- Before the change, a binary under `node_modules/lefthook-darwin-arm64/bin/` wrote its absolute path; now it writes `lefthook run validate`, and `lefthook: pnpm exec lefthook` still produces `pnpm exec lefthook run validate`.


internal/command/install_ai.go
```diff
@@ -33,42 +33,43 @@ const (
...
 
-func lefthookRunCommand(bin, hookName string, quoteBin bool) string {
-	if quoteBin {
-		bin = shellQuotePath(bin)
+// portableLefthookBin maps the path of the running lefthook executable to a
+// command that works on any machine. A binary installed from the npm package
+// lives under node_modules, which is not on PATH, so it is called via npx.
+func portableLefthookBin(exe string) string {
+	if slices.Contains(strings.Split(filepath.ToSlash(exe), "/"), "node_modules") {
+		return npxLefthookBin
```
mrexox: Maybe it should be better just to rely on globally installed lefthook, rather than `npx`. I guess some users may want to use the lefthook binary from their repo, some will use `pnpm` for managing packages, and `npx` will download the lefthook for them.
What do you thin about having just `lefthook` here?

anandghegde: Agreed, switched to plain `lefthook` in 0d70824. The `npx` guess was too narrow anyway (pnpm/yarn/bun users), and anyone running lefthook from a package manager can already set `lefthook: pnpm exec lefthook` etc., which is still used as is. Docs and the PR description are updated to match.

internal/command/install_ai.go
```diff
@@ -37,40 +37,24 @@ const (
...
 		}
 	}
 
-	exe, err := os.Executable()
-	if err != nil {
-		return lefthookBinName, false
-	}
-
-	return filepath.Clean(exe), true
+	return lefthookBinName
```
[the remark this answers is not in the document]
anandghegde: The npx fallback was in this PR's first commit and was dropped deliberately after the maintainer review above: `npx` guesses wrong for pnpm/yarn/bun users and can download lefthook on the fly. On `master` there was no portable fallback either. It wrote the absolute `node_modules/.../bin/lefthook` path, which only works on the machine that ran install and breaks after upgrades (#1482).

So a project that has lefthook only as a package dependency and no global binary now needs to set `lefthook: npx lefthook` / `pnpm exec lefthook`. That value is still used as is, and `docs/configuration/ai.md` now says so. The git hook shim can fall back through `node_modules` because it's a generated script, but these commands are single lines in committed JSON, so leaving the choice to that config option seems cleaner than embedding a guess. No code change.

PR #1547, opened 2026-09-16
feat: propagate forced colors to hook commands via CLICOLOR_FORCE

Yash121l:
Follow-up to #1538, as promised there.

### Context

#1538 made `--colors on` and `colors: true` paint lefthook's own output when stdout is not a terminal. The setting never reached the commands lefthook runs. The executor propagated colour intent in one direction only:

```go
if e.logger.NoColors() {
	envs = append(envs, "NO_COLOR=true")
}
```

So `--colors off` reached eslint, prettier and cargo, but `--colors on` did not. Lefthook captures job output through a pipe whenever there is no pty, so those tools detect a non-terminal and print plain text. In CI the lefthook header is coloured while every linter under it is not.

`CLICOLOR_FORCE` is already documented as forcing colours "in lefthook and all subcommands", so this makes the flag and the config option match what the env var has always done.

### Changes

- `Logger.ColorsForced()`, the counterpart to `NoColors()`. True only when colours were set explicitly and not disabled, so `auto` is unaffected.
- The colour env decision moves out of the unix and windows executors, which had the same block, into one `colorEnv()` helper in `executor.go`. It returns `NO_COLOR=true` when colours are off, `CLICOLOR_FORCE=1` when they are forced, and nothing otherwise.
- A `CLICOLOR_FORCE` already set in the job's `env:` or in lefthook's own environment is left alone. Someone who sets `CLICOLOR_FORCE=0` means it, and since the child env is `append(os.Environ(), envs...)` with the last entry winning, appending unconditionally would silently override them.
- One paragraph in `docs/configuration/colors.md`.

### Verification

`TestExecute_Colors` is a map-based table test that runs `echo "[$CLICOLOR_FORCE]"` through the real executor and checks what the command saw: forced colours give `[1]`, off and auto give `[]`, and a `CLICOLOR_FORCE=0` from the job env or the process env survives as `[0]`. The first case fails on current master and passes with this change.

The built binary under `CI=true TERM=dumb` with output piped: `--colors on` gives the job `CLICOLOR_FORCE=[1] NO_COLOR=[]`, `--colors off` gives `CLICOLOR_FORCE=[] NO_COLOR=[true]`, the default gives neither.

`go test -race ./...` and `go vet ./...` pass.

### PR checklist

- [ ] `make lint` passes: the pinned golangci-lint 2.11.4 panics on Go 1.27 locally (`file requires newer Go version go1.27`, same on master). golangci-lint 2.13.2 with the repo's config reports nothing in the touched files; CI's pinned job is the real check.
- [x] `make test` passes
- [x] Docs updated if behavior changed or new config option added


mrexox:
Thank you!

